PR #39, opened 2016-04-29
Swapped zcat for gzcat, as zcat is very much broken on macos

swingingsimian:
Title says it all really. zcat on macos adds '.Z' to the file name automatically, meaning it can't see the file. Bismark, currently doesn't catch this error and carries on regardless.

Thanks

Nathan


swingingsimian:
Sorry this, should be gunzip -c for cross platform compatibility, will amend and issue another pull request

PR #40, opened 2016-04-29
Swapped zcat for gunzip -c, as zcat is very much broken on macos

swingingsimian:
Title says it all really. zcat on macos adds '.Z' to the file name automatically, meaning it can't see the file. Bismark, currently doesn't catch this error and carries on regardless.

Thanks

Nathan


FelixKrueger:
Thanks for this, I am sure this will be helpful for the Mac community!

PR #66, opened 2016-10-07
Pandoc HTML docs

ewels:
Made the HTML docs output with pandoc, with some customisation to make it pretty.

New workflow:

1. Edit the markdown file - now in `Docs/README.md`
2. Run the script to create the HTML file: `bash make_docs.sh`
3. Commit and push

We should be able to automate this in the future.


ewels:
Also, note that I rewrote the main repo readme a bit. Hope that's ok. Say if you want me to change anything.

PR #70, opened 2016-10-21
Wrote Perl Markdown - HTML converter.

ewels:
New docs creater thingie! Perl script that takes the documentation markdown and converts it to a nice HTML page.

Majority of the code is to add `id` tags to the headings and then build a nav / table of contents to link to those headings. It's a bit nasty, I wrote it in a rush on a train, so it can probably be tidied up a little. Seems to work though.

Phil


FelixKrueger:
Awesome, great job! And all that during a DDoS attack!

PR #77, opened 2016-11-18
Update bismark 

Shicheng-Guo:
more detail about seed length setting.


FelixKrueger:
Thanks for the suggestions, can't merge because of conflicts though. Will need to look on Monday.

Shicheng-Guo:
I am not familiar with github. I should keep update as you. Now I now how to do it.

PR #80, opened 2016-12-02
Merge pull request #1 from FelixKrueger/master

Shicheng-Guo:
update from origin


Shicheng-Guo:
location of time tramp is corrected. 

image

PR #83, opened 2016-12-05
add minimum mapping quality for bismark_methylation_extractor

Shicheng-Guo:
add minimum mapping quality for bismark_methylation_extractor

--mapq

### DIFF
--- a/bismark_methylation_extractor
+++ b/bismark_methylation_extractor
@@ -979,6 +979,7 @@ sub process_commandline{
 				 'ignore_3prime=i'      => \$ignore_3prime,
 				 'ignore_3prime_r2=i'   => \$ignore_3prime_r2,
 				 'multicore=i'          => \$multicore,
+				 'mapq=i'           => \$mapq;
 	);
 
   ### EXIT ON ERROR if there were errors with any of the supplied options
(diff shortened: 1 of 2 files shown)


bismark_methylation_extractor
```diff
@@ -1123,7 +1124,7 @@ VERSION
 	    open (DETERMINE,"gunzip -c $file |") or die "Unable to read from gzipped file $file: $!\n";
 	}
 	elsif ($file =~ /\.cram$/ || $file =~ /\.bam$/ || isBam($file) ){ ### this would allow to read BAM files that do not end in *.bam
-	    open (DETERMINE,"$samtools_path view -h $file |") or die "Unable to read from BAM/CRAM file $file: $!\n";
+	    open (DETERMINE,"$samtools_path view -q $mapq -h $file |") or die "Unable to read from BAM/CRAM file $file: $!\n";
```
FelixKrueger: A couple of things: 
- the DETERMINE file handle is just the one where single-end or paired-end files are detected. It would have to changed at several more locations.
- this addition break the code in cases where --mapq was not specified. This would need looking into.

- and most importantly: Filtering on MAPQ is probably not what you would want to do anyway. Since Bismark is only allowing unique alignments the MAPQ score does not reflect the ambiguity of the read but it is simply a mix between the score_min setting (which are very strict by default) and the number of mismatches and/or gaps in the alignment. Thus, filtering on MAPQ would more or less be the same as filtering on mismatches/gaps, which can and should be done during the mapping step.

PR #102, opened 2017-04-26
update

Shicheng-Guo:


FelixKrueger:
Hmm, no description? I think I have replied to most of the changes in previous issues and/or pull requests already.

PR #115, opened 2017-06-28
Added --dovetail back for backwards compatibility.

ewels:
This is kind of untested. But how badly can it break things? 😈


FelixKrueger:
I was hoping you would simply change the `--help` text, but I guess this will work as well...

ewels:
Yeah, should probably change the `--help` text as well, seeing as this flag doesn't actually have any effect.

PR #126, opened 2017-09-21
A couple of small fixes

rifius:


rifius:
After processing all recorded positions, CpG report file gets cytosines for uncovered positions of as-yet-unseen chromosomes. Will try to track down and suggest fix.

FelixKrueger:
Hi Carlos,

I have now changed the NOMe-Seq processing so that it simply doesn't process chromosomes that were not covered in the experiment at all. See [here](b5fe29d6feb9a8e7d4769d7d86b408ceeddfa8e3).

PR #128, opened 2017-09-27
Fixed file handling with gzip and split by chromosome yes/no

rifius:
in CpG and GpC reports.
Not very happy, the position starting at zero logic could be simplified a bit to reduce clutter, and is not considered in the GpC report.  Did not open an issue either.

### DIFF
--- a/coverage2cytosine
+++ b/coverage2cytosine
@@ -99,7 +99,7 @@ generate_genome_wide_cytosine_report($coverage_infile);
 # which recommends this merging process to increase coverage per CpG and reduce the memory burden for its processing
 # Merge CpGs does not work in conjuction with --nome-seq (since some positions are filtered out).
 
-if ($merge_CpGs){
+if ($merge_CpGs) {
     # we only allow this operation if the report is limited to CpG context, and for a single report for the entire genome for the time being
     combine_CpGs_to_single_CG_entity($cytosine_out);
 }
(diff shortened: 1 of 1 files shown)


rifius:
Done.  #129 opened and linked.

FelixKrueger:
Hi Carlos,

Thanks for your suggestions to tidy up the file handling. I have now merged the pull request even though this proved to have been a little premature as it was only only working when `--NOMe-seq` had been specified. I have now fixed the problems (in this commit 74e2caeb3429c8e93023f7a120ce89c26866a747) so it is now working with any (permitted) combination of normal, `--gzip`, `--gc`, `--NOMe` and `--CX_context`.

Unfortunately however the option `--split_by_chromosome` is now no longer working in any mode. Once I find some more time and got my head round your code referencing I will try to look at this. Right now I need to attend my birthday steakhouse dinner though, first things first! :smiley_cat:

rifius:
Hi Felix,

Happy birthday!  A week ago I had mine in a Brazilian barbecue, an almost
epicurean debauchery.
I'll try to check your commit and suggest another pull.  The idea (I
thought) was simple:  work out the name of the file in a function, calling
it on the first opportunity to actually write anything to file.

Cheers,
--c
--



*Dr Carlos Riveros*

HMRI Bioinformatics Fellow

*P* (02) 4042 0478
*E* Carlos.Riveros@hmri.org.au <carlos.riveros@hmri.org.au>

*W* hmri.org.au

FelixKrueger:
Hi rifius,

Now you got me intrigued what happened at your Brazilian BBQ, but happy birthday as well!
The idea to clean up file names is certainly a good one, now we just need to get it working again (we don't ever use `--split_by_chromosome` anyway...). In that case I will just wait another day or so for a pull request, else I'll fix it myself later this week.

PR #195, opened 2018-08-16
update my local version to Felix lastest version

Shicheng-Guo:


FelixKrueger:
Do I need to do something about this?

PR #203, opened 2018-09-19
Add new parameter bowtie_threads to parallelize bowtie index creation

rakarnik:


FelixKrueger:
Thanks for this Rahul, I will add some documentation and make sure it doesn't conflict with the `--bowtie1` option (since Bowtie1 does not support multi-threading).

FelixKrueger:
I did a few tests using the human genome today (GRCh38), and it does indeed work nicely. Here are the results (roughly) for the genome preparation:

```
number of cores      wall clock time
====================================
single core             3h 30min
--bowtie_threads 2      2h 50min
--bowtie_threads 4      1h 55min
--bowtie_threads 8      1h 22min
```

Thanks again for getting this rolling.

rakarnik:
Thanks Felix! I ran it with 20 threads here, and saw a nice speed up too. Good point about bow tie1 — maybe you can name the option bowtie2_threads to make it clearer, in addition to handling —bowtie1 and —bowtie2_threads in the code.

FelixKrueger:
I just changed the option to `--parallel` to be in line with the option in `bismark` and `bismark_methylation_extractor`. Also added some more text for `--help` to warn users that the option will use twice as many resources as the value the they specify...

rakarnik:
Sounds good! By the way, it looks like Bowtie1 supports the --threads argument too. I am using Bowtie v1.2.1.1.

FelixKrueger:
Thanks for the comment, I have now updated Bowtie 1 on our cluster, and also updated `bismark_genome_preparation` to allow `--parallel` for the Bowtie1 indexing process.

PR #240, opened 2019-03-13
Bc ignore read id descriptions

### LINKED ISSUE #236: Concatenating optional read ID descriptions to read ID's in outputted BAM
In order "to avoid problems with truncated read ID when they contain white spaces", whitespace in the the read ID line (Line 1's of the FASTQ format) if currently being replaced underscores.

https://github.com/FelixKrueger/Bismark/blob/782b3e543842a9c40e435a19fbe4660f22020aa8/bismark#L5885

However, read ID's can be followed by an optional description that is currently being concatenated to the read ID in the outputted BAM/SAM. This can cause the output data from Bismarck to be altered/differ from the input FASTQ's and the output of other tools, such as picard tools or gatk tools, which ignore the read ID optional description.

[FASTQ Format:](https://en.wikipedia.org/wiki/FASTQ_format)
A FASTQ file normally uses four lines per sequence.

**Line 1 begins with a '@' character and is followed by a sequence identifier and an optional description (like a FASTA title line).**
Line 2 is the raw sequence letters.
Line 3 begins with a '+' character and is optionally followed by the same sequence identifier (and any description) again.
Line 4 encodes the quality values for the sequence in Line 2, and must contain the same number of symbols as letters in the sequence.

benjamincarlin:
Fixes #236 

@FelixKrueger thank you for your time! I have been exploring possible solutions for this issue and decided to make a PR for updating the fix_id method. Please let me know if this solution sounds okay or if there is anything else I can do!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,8 @@ script:
   - "./bismark --non_directional --genome ./travis_files/ -1 ./travis_files/test_R1.fastq.gz -2 ./travis_files/test_R2.fastq.gz"
   - "./bismark --hisat2 --pbat --genome ./travis_files/ -1 ./travis_files/test_R1.fastq.gz -2 ./travis_files/test_R2.fastq.gz"
   - "./bismark --hisat2 --non_directional --genome ./travis_files/ -1 ./travis_files/test_R1.fastq.gz -2 ./travis_files/test_R2.fastq.gz"
-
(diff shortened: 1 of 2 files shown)


FelixKrueger:
Great, now we have both done the same thing :)

benjamincarlin:
sorry just saw this! I apologize for opening the PR/Issue again. Please ignore!

Thanks again!

FelixKrueger:
Thanks to you, too!

PR #246, opened 2019-04-01
Added soft-clipping (i.e. local alignments) support for Bowtie 2 alignments 

FelixKrueger:
This entailed adapting the alignment mode in Bismark, and required minor changes to `deduplicate_bismark`, `bismark_methylation_extractor` and `bam2nuc`.


FelixKrueger:
Need to ensure that it also works for HISAT2 alignments.

PR #250, opened 2019-04-13
append hisat2 to .bam output file names

phue:
Currently, the hisat2 aligner information is not contained in the file name when using `--bam` _if_ `--multicore` is set.
This causes `bismark2summary` to fail because the aligner can not be detected. 
This commit should fix it by appending hisat2 to the output filename.


FelixKrueger:
Excellent, many thanks.

PR #282, opened 2019-10-17
fix scMin calculation in calc_map

### LINKED ISSUE #281: score_min not correctly computed in calc_map, leading to wrong MAPQ output
Hi Felix,
thanks a lot for all your work and for sharing so much of your experience through qcfail.com etc.

I noticed that the recomputed mapping qualities don't seem to be correct when using local alignments. I saw #260 and also that you included the updated code for computing the MAPQ in the latest release from today.

However, I think that the calculation of scMin also needs to be updated to account for the logarithmic formula (min_score=G,...,...)

This is at the start of sub calc_mapq, in line 3797:

```perl
	my $scMin = $score_min_intercept + $score_min_slope * $read1Len;
	### read2Len is only defined for paired-end reads, so for single-end mode we can just a score min value for read 1
	if (defined $read2Len){
		$scMin   += $score_min_intercept + $score_min_slope * $read2Len;
	}
```
I tested your new mapq calculation code, and while it doesn't produce 0 and 2 as before, it mainly produces MAPQ 22 for valid alignments, which is in line with a wrong scMin calculation.

Does that make sense? Should I submit a pull request with my fix?

Thank you,
Stephen

stephenkraemer:
closes #281 

scMin calculation is either c + a * read_length or c + a * ln(read_length) depending on whether alignment happens in local mode

### DIFF
--- a/bismark
+++ b/bismark
@@ -3797,10 +3797,16 @@ sub determine_number_of_transliterations_performed{
 sub calc_mapq {
 	my ($read1Len, $read2Len, $AS_best, $AS_secBest) = @_;
 
-	my $scMin = $score_min_intercept + $score_min_slope * $read1Len;
+	# Calculate the minimum alignment score either with linear or logarithmic function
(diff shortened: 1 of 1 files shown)


FelixKrueger:
Many thanks for this. Cheers, Felix

PR #379, opened 2020-10-02
Finished moving CI tests from Travis to Github Actions

FelixKrueger:


ewels:
It upsets me slightly that the test data is still in a directory called `./travis_files/` 😅

One of the things that I quite like about GHA is that you can hide all of this stuff away into `.github` and not pollute the project root directory.

FelixKrueger:
I get that. I was considering it as well. I am considering moving it :)

Issue #494, opened 2022-06-10
Bismark outputs corrupted result files if one of child process fails


iromeo:
Hi, I'm using Bismark (0.23.0) in Snakemake pipeline to process many of the WGBS. I've noticed that some of my BAMs have fewer reads than expected, and ambiguous/unaligned FASTQ files have invalid format. I run my bismark in `multicore` mode. When I checked logs, I noticed that some of the cores could fail `Child process terminated with exit signal: '13'` but main snakemake process continuous working and return 0 exit code, so I get an incorrect result which hard to notice in batch processing mode. In this case Bismark should fail with an error. Child cores failed likely due to cluster issues (e.g. lack of HDD space or etc), I successfully re-aligned one of the problematic samples and got correct results and files.

Some details:
* Command to launch Bismark: `bismark -X 600 --gzip --multicore 4 --ambiguous --unmapped --bowtie2 $(dirname resources/indexes/hg38/Bisulfite_Genome) -1 data/reads/wgbs_y23_pooled/Clean/FE04/FE04_1.fq.gz -2 data/reads/wgbs_y23_pooled/Clean/FE04/FE04_2.fq.gz`
* Interesting pars of log:
```
Current dir: /scratch1/fs1/martyomov/rcherniatchik/aging-wgbs-smk-pipeline-v3/.snakemake/shadow/tmp60xczd7h
Bowtie 2 seems to be working fine (tested command 'bowtie2 --version' [2.4.2])
Output format is BAM (default)
Alignments will be written out in BAM format. Samtools found here: '/scratch1/fs1/martyomov/rcherniatchik/aging-wgbs-smk-pipeline-v3/.snakemake/conda/af71104ea71ba452675d70f85d913ba3/bin/samtools'
Reference genome folder provided is resources/indexes/hg38/     (absolute path is '/storage1/fs1/martyomov/Active/projects/aging/LongitudinalYear123/indexes/hg38/)'
FastQ format assumed (by default)

Input files to be analysed (in current folder '/scratch1/fs1/martyomov/rcherniatchik/aging-wgbs-smk-pipeline-v3/.snakemake/shadow/tmp60xczd7h'):
data/reads/wgbs_y23_pooled/Clean/FE04/FE04_1.fq.gz
data/reads/wgbs_y23_pooled/Clean/FE04/FE04_2.fq.gz
Library is assumed to be strand-specific (directional), alignments to strands complementary to the original top or bottom strands will be ignored (i.e. not performed!)
[...]

PR #497, opened 2022-06-14
Fix for #494: Multicore Bismark corrupts result files if one of child process fails

iromeo:
The suggested fix for #494: Just stop execution if not all cores have finished successfully


FelixKrueger:
Killing the parallel/multicore run if one or more child processes have not completed successfully (e.g. lack of memory, disk space etc).

PR #506, opened 2022-07-12
Testing: Fix CI tests

semenko:
These force the CI tests to use Python 3.9 and obtain a recent-ish version of bowtie2.

Tests still fail, but they fail a bit later on now. 😛


semenko:
[wrong branch merge -- reopened as #507]

PR #544, opened 2022-11-10
Tidied up some usage docs.

ewels:
Mentioned that summary only needs BAM filenames and not BAMs.


FelixKrueger:
I should probably mention that I have used a `for` loop and `touch` to re-generate the BAM files to emulate the behaviour you are describing for `bismark2summary`....

ewels:
> I should probably mention that I have used a `for` loop and `touch` to re-generate the BAM files to emulate the behaviour you are describing for `bismark2summary`....

How come? Just passing the filenames to the script seems to work fine?

FelixKrueger:
This was outside of Nextflow, when the original BAM files were not actually available for some reason. So passing the filenames, or doing `touch` to re-generate dummy files both seem to do the job

PR #720, opened 2024-12-11
Tab formatting

ahanden:
For ease of editing, I fixed all the indentation in bismark to be consistent. Indentation is now two spaces.


FelixKrueger:
Thanks for your positive intent, but I have to say I am not a fan of 2 space formatting (and if we wanted to make a push for consistent formatting it should consistent across all modules).

Issue #798, opened 2026-05-23
epic(extractor): port bismark_methylation_extractor to Rust


FelixKrueger:
## Summary

Port the Perl `bismark_methylation_extractor` (~6,050 LOC) to a Rust binary. By far the biggest single-tool rewrite in the suite: it dominates pipeline runtime (16%, with superlinear scaling on real WGBS data), drives downstream bedGraph + cytosine_report outputs, and has the broadest flag surface of any downstream tool.

## Change

New crate `bismark-extractor` in the workspace. Library+binary split — pure logic lives in `lib.rs` (testable), `main.rs` is thin orchestration.

CLI matches Perl `bismark_methylation_extractor`:

- Mode flags: `--single` / `--paired`, `--comprehensive`, `--merge_non_CpG`, `--CX`, `--yacht`
- Read-region flags: `--ignore`, `--ignore_3prime`, `--ignore_r2`, `--ignore_3prime_r2`
- Overlap handling: `--no_overlap` (default on), `--include_overlap`
- M-bias: `--mbias_off`, `--mbias_only`
- Output: `--gzip`, `--bedGraph`, `--cytosine_report`, `--genome <path>`, `--output <dir>`
- Parallelism: `--multicore N` (rayon-based) and `--parallel` alias

## Implementation notes — structural design choices

These choices are deliberate prevention of known Perl-to-Rust port failure modes:

1. **`BismarkStrand` decided ONCE per read pair**, threaded through extraction. The strand is a property of the pair, not of an individual call — making it structurally impossible to split one read's calls across multiple strand-specific output files.
2. **M-bias counters per (context × read identity)**: 6 tables — `(CpG, R1)`, `(CHG, R1)`, `(CHH, R1)`, `(CpG, R2)`, `(CHG, R2)`, `(CHH, R2)`. Stored as `[MbiasTable; 2]` indexed by read_identity, with explicit per-context iteration in the writer.
3. **CHG and CHH context handling is first-class.** Every code path that handles CpG must also handle CHG and CHH explicitly. No `match CytosineContext { CpG => ..., _ => {} }` shortcuts. Tested via a fixture with measurable CHG and CHH signal — synthetic small-genome data doesn't exercise this.
4. **Argument structs, not 14-parameter functions.** `extract_calls` and `process_pair` take typed parameter structs (`ExtractParams`, `PairParams`) — makes argument routing structurally verifiable.
[...]

PR #883, opened 2026-05-29
feat(extractor): console progress/diagnostics logging (stderr, --quiet)

### LINKED ISSUE #882: feat(extractor): console progress/diagnostics logging (Perl-equivalent, stderr, --quiet)
## Summary

The Rust `bismark-methylation-extractor-rs` is nearly silent — it prints only the `contains data -> kept` / `was empty -> deleted` cleanup lines. The Perl `bismark_methylation_extractor` prints a rich diagnostic log that users rely on for debugging (banner, detected SE/PE mode, parameter summary, a `Processed lines: N` progress counter every 500k, and a final per-context methylation summary). On a long run, Rust's silence is indistinguishable from a hang, and silent SE/PE auto-detection hides wrong-guesses.

**Goal:** give the Rust port Perl-equivalent observability.

## Not on the beta.1 byte-identity critical path

The Phase H matrix compares **output files** (`cmp` on reports/M-bias, sorted-md5 on data files) and only `tail -3`s the console for display — stdout/stderr is outside the byte-identity invariant (verified against `phase_h_smoke.sh` / `phase_h_pe_matrix.sh`). So this can land post-`v1.0.0-beta.1` without reopening the gate.

## Informed design decisions (from Perl/Rust investigation)

- **Stream → stderr.** Perl sends *all* diagnostics to stderr via `warn` (232 calls; 0 `print STDOUT`). The Rust port already emits kept/deleted via `eprintln!` → stderr. Keep that model; stdout stays clean. `--version` stays on stdout.
- **Default on, add `-q`/`--quiet`** to suppress the informational log (errors/warnings still print).
- **No `--verbose` noise tier needed** — the Perl per-`@SQ` "skipping SAM header line" dump is an artifact of Perl's line-by-line SAM parsing; the Rust port reads the header structurally via noodles, so that noise doesn't exist.
- **Final summary is a near-free mirror** — those numbers already exist (computed for the byte-identical `_splitting_report.txt`).
- **No new dependency** — raw `eprintln!` via a small quiet-gate helper.

## Lines to add (stderr)

1. Version banner
2. Mode detection (`Treating file(s) as {single|paired}-end data …`)
3. Parameter summary (cores, output dir, ignore_*, overlap mode, comprehensive/merge/gzip)
4. `Processed lines: N` every 500k
5. Final per-context methylation summary

## Dual-dispatch caveat

The read loop exists in BOTH `route.rs` (`--parallel 1`) and `parallel.rs` (`--parallel N`); the #876/#879 fixes both had to touch both. The progress counter must be added to both paths or it silently disappears at one parallelism setting.

## Plan

`plans/05262026_bismark-extractor/CONSOLE_LOGGING_PLAN.md` (full behavior, implementation outline, edge cases, verification incl. a `phase_h_smoke.sh` byte-identity regression check).

Part of epic #798. Sibling of #878 (CI/observability follow-ups).

FelixKrueger:
## Summary

Adds Perl-equivalent console diagnostics to `bismark-methylation-extractor-rs`, which was nearly silent (only the `kept`/`deleted` cleanup lines). All output goes to **stderr** (matching Perl's `warn` model — 232 `warn`s, 0 `print STDOUT`); stdout stays clean. Closes #882.

New output (default on; `-q`/`--quiet` silences info, warnings/errors still print):
- Startup banner (`CARGO_PKG_VERSION`, not the v0.25.1-locked `BISMARK_VERSION`)
- SE/PE mode + parameter summary (workers, output dir, ignore_*, overlap, gzip)
- `@HD` + `@PG` **header provenance** (Bismark version + command line, samtools chain) — `--verbose` adds the `@SQ` contig dictionary, which is dropped by default as noise
- `Processed lines: N` every 500k
- Final per-context methylation summary

## Why this is safe for the byte-identity gate

Console output is **outside** the Phase H byte-identity contract — the matrix compares output *files* (`cmp` on reports/M-bias, sorted-md5 on data files) and only `tail -3`s the console for display. No output-file bytes change, so the tagged `v1.0.0-beta.1` behaviour is unaffected. (To be confirmed by a `phase_h_smoke.sh` SE+PE regression run on colossal.)

## Design notes

- New `src/logging.rs`: `Logger { quiet, verbose }` (Copy) with pure, unit-tested text builders.
- Progress ticks at the **single live read site** (`parallel.rs::producer_loop`), counting each SAM record (+1 SE, +2 PE) — a plain local counter (single producer thread, no atomic) that matches Perl's `line_count` byte-for-byte.
- Header provenance reuses `detect_paired_from_header`'s serialize-to-SAM-text idiom (robust to noodles tag-reordering); implemented locally in the extractor (already deps `noodles-sam`) to avoid a `bismark-io` `=`-pin version cascade.
- `@HD`/`@PG` kept, `@SQ` dropped by default (per the in-review distinction: `@PG` provenance is valuable, the 190+ `@SQ` contig dump is noise).

## Two-counter correctness (caught in code review)

`Processed N lines in total` uses `records_processed` (Perl `sequences_count`, `:2479`/`:2482` = pairs for PE); `methylation call strings processed` uses `call_strings_processed` (2×pairs). These differ for PE; SE they're equal. Mirrors the byte-identical `_splitting_report.txt`.

## Verification

- 101 lib tests (3 new `logging` unit tests: quiet gate, summary shape+%, `@SQ`-drop) + all integration binaries pass; `clippy --all-targets -D warnings` clean.
- Dual plan-review + dual code-review + plan-manager (COMPLETE). Reports under `plans/05262026_bismark-extractor/CONSOLE_LOGGING_*`.
- **Pending before merge**: colossal eyeball of the real output + `phase_h_smoke.sh` SE/PE/AutoDetect byte-identity regression.

closes #882 · refs #798

### DIFF
--- a/plans/05262026_bismark-extractor/CONSOLE_LOGGING_CODE_REVIEW_A.md
+++ b/plans/05262026_bismark-extractor/CONSOLE_LOGGING_CODE_REVIEW_A.md
@@ -1,0 +1,206 @@
+# Code Review A — extractor console progress/diagnostics logging (commit `52b05f8`)
+
+**Reviewer:** A (fresh context)
+**Branch:** `feat-extractor-console-logging`
+**Commit:** `52b05f80781852f358fc591d582d8ec88879d1e5`
+**Plan:** `plans/05262026_bismark-extractor/CONSOLE_LOGGING_PLAN.md`
+**Verdict:** APPROVE with one fix applied (PE counter mismatch) + minor recommendations.
+
+---
+
+## Summary
+
+The change is well-scoped, faithful to the (rev-1) plan, and the documented
+deviations are sound. New `src/logging.rs` introduces a `Copy` `Logger { quiet,
+verbose }` with pure, testable text builders; the single live read site
+(`parallel.rs::producer_loop`) ticks `Processed lines: N` every 500k counting
++1/SE record and +2/PE pair; banner/params/provenance fire exactly once on the
+main thread before the reader is moved into the producer. The `--quiet` gate is
+applied to every informational line and the genuine warnings/errors are left
+ungated. Nothing in the diff touches output **file** content, so the beta.1
+byte-identity gate is unaffected.
+
+I found and **fixed one correctness bug**: the console final-summary
+`"Processed N lines in total"` line used `call_strings_processed` instead of
+`records_processed`, so for PE inputs the console showed 2×pairs where the
+`_splitting_report.txt` file (and Perl) show pairs. Details below.
+
+`cargo test -p bismark-extractor` (incl. the 3 new logging tests) and
+`cargo clippy --all-targets -D warnings` are green after the fix.
+
+---
+
+## Issues by area
+
+### 1. Logic / correctness
+
+**[FIXED — High] PE `"Processed N lines in total"` used the wrong counter.**
+`final_summary_text` (`logging.rs:203-217`) used `report.call_strings_processed`
+for *both* the `"Processed N lines in total"` line and the `"...call strings
+processed: N"` line. But the actual `_splitting_report.txt` writer
+(`output.rs:683`) and Perl `:2482` (`$sequences_count`) use `records_processed`
+for the first line:
+- `output.rs:683` → `Processed {records_processed} lines in total`
+- `output.rs:690` → `Total number of methylation call strings processed:
+  {call_strings_processed}`
+
+For SE these counters are equal (no visible effect), but for PE
+`records_processed` = pairs while `call_strings_processed` = 2×pairs
+(see the `SplittingReport` doc comments at `output.rs:403-420`). So on any PE
+run the console summary's first line would have shown **double** the file's
+value — a direct contradiction of the plan's "mirror of splitting_report"
+requirement and of Perl's stderr `warn` output. Fixed by binding `lines =
+records_processed` and adding a separate `call_strings` binding; the existing
+unit test was strengthened to give the two counters distinct values so the
+distinction is locked.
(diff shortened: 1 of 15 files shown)


FelixKrueger:
## Stage 2 verification on colossal — PASS ✅

Ran the full Stage-2 gate against the feature branch on colossal (10M SE + PE, Perl Bismark v0.25.1 reference).

**Byte-identity (output files unchanged — the key guard):**
- SE smoke (`directional_10M`): **PASS** — 8/8 files raw-identical.
- PE smoke (`SRR24827378_10M`, auto-detect): **PASS** — 8/8 files, incl. `M-bias.txt` **11,443 B** + `_splitting_report.txt` **875 B** at their locked baselines.

Confirms console-logging is presentation-only; output files are bit-for-bit unchanged. (Console output is outside the Phase H byte-identity contract — the matrix compares files, not stderr.)

**Console behaviour (eyeball):**
- Default: banner + SE/PE mode + param summary + `@HD`/`@PG` provenance (no `@SQ` flood) + `Processed lines: N` every 500k + final summary. Final-summary numbers match the Perl reference exactly (8501508 lines, 105567573 C's, 81.8/2.6/1.2%).
- `--quiet`: **0** info lines, but a forced error still prints (`error: input file does not exist`).
- `--verbose`: adds the 194 `@SQ` lines.
- `--parallel 10`: progress counter ticks cleanly (single producer site).

Verified on commit `e43662f`; the later `73c3346` (mode-line parenthetical) is a stderr-string-only delta that cannot affect output files, so the verdict carries.

Issue #884, opened 2026-05-29
perf(extractor): parallel-I/O redesign — batch records + worker-side output (anti-scales at N>1)


FelixKrueger:
## Problem

The Rust extractor **anti-scales** at `--parallel N>1`: N=1 is ~6× faster than Perl single-core (~90 s vs ~560 s on 10M PE), but N=4 ≈ Perl-N=4 and is *slower than Rust-N=1* (~170 s). Perl's fork model scales ~3.5×; ours regresses. (Follow-up to #876 Finding #4.)

## Spike findings

Full report: `plans/05262026_bismark-extractor/spikes/SPIKE_parallel_io.md`.

**1. Parallel BGZF decode scales — but is NOT the bottleneck.** Bench (10M PE, 1.2 GB, 15.4M records, colossal), noodles `MultithreadedReader`:

| workers | 1 | 2 | 4 | 8 | 16 |
|---|---|---|---|---|---|
| speedup vs single | 0.69× | 1.37× | 2.88× | 5.71× | 8.65× |

Corroborated by `samtools view -@`: ~6.7× at 7 threads, knee at ~8. **But decode is only ~6.8 s of the ~90 s** N=1 run → wiring `MultithreadedReader` alone ≈ 7%.

**2. The real wall = the single-threaded collector.** The ~83 s remainder is per-record extraction + the collector writing 5+ GB and reordering **15.4M records one-at-a-time through a `BTreeMap`**. N=4 anti-scaling = this serial tail + per-record channel/BTreeMap overhead overwhelming the cheap parallel `extract_calls` (Amdahl).

**3. Trim Galore Oxidised Edition (Rust v2.2.0) already solved this** (`github.com/FelixKrueger/TrimGalore`): it *abandoned* the "single main → gzip writer" model (gzip was ~60% of runtime through one thread) for **per-worker compression + ordered multi-member concatenation**, batching ~4096 records per channel message. Our collector IS the model TG-OE dropped.

## Proposed direction (profile-first, Phase-H-gated)

1. **Profile** the ~83 s into extract / route / write before committing (size levers with data).
2. **Batch records** (~4096) per channel message with a `u64` seq — cut per-record coordination ~4000× (likely the anti-scaling root cause; std-only).
[...]

PR #888, opened 2026-05-29
perf(extractor): parallel --gzip via gzp ParCompress in the collector (#884 R2)

FelixKrueger:
## Summary
R2 of #884: parallelize `--gzip` split-file compression by swapping the single-threaded `flate2::write::GzEncoder` writers for **gzp's `ParCompress<Gzip>`** (parallel gzip, `deflate_rust` pure-Rust backend, no cmake).

The Phase-0 spike decomposed the `--gzip` cost (10M PE): extract/decode floor ~16.7s, `.txt` write ~6s, and a **~52s single-threaded gzip compression wall**. gzp parallelizes that wall on its own pool. Chosen over the invasive worker-side-members design in the R2 plan: **both independent plan reviewers recommended gzp-in-collector (ALT-1)** and a throwaway spike confirmed it (`.gz` 75s → 18s).

## What changed
- **`output.rs::open_writer`**: the `--gzip` arm now builds `gzp::par::compress::ParCompress<gzp::deflate::Gzip>`. New `GZIP_COMPRESS_THREADS = 4` const, **decoupled from `--parallel`** so the common `--parallel 1 --gzip` default still gets the win (tying it to `--parallel` would leave the default single-threaded ~75s).
- **Single-member gzip**: gzp's `Gzip` format emits one header + sync-flushed DEFLATE blocks + one stream-wide CRC/ISIZE footer (verified in gzp source + a multibatch test). Existing single-member `GzDecoder` tests stay valid — **no `MultiGzDecoder` migration**. Decompressed content is byte-identical; only raw `.gz` framing differs. Collector ordering, version header, empty-sweep, and the plain `.txt` path are untouched.
- **`Cargo.toml`**: add `gzp = "=0.11.3"` (`deflate_rust`); move `flate2` to `[dev-dependencies]` (now test-only — tests decode `.gz` via `GzDecoder`).
- **Test**: `parallel_gzip_multibatch_decompresses_identical_across_n_and_to_plain` (8199 records > 2×BATCH_SIZE) guards single-member framing + N=1≡N=4 + identity-to-plain.

## Performance (colossal, 10M, `--parallel 4`)
| mode | Perl | Rust | Speedup |
|---|---|---|---|
| SE `--gzip` | 72s | 8s | 9.0× |
| PE `--gzip` | 167s | 19s | 8.7× |

(≥4× SPEC §9.7 target cleared.)

## Verification
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p bismark-extractor` (102+ tests) ✅
- **Colossal `--gzip` SE+PE `phase_h_smoke` byte-identity: PASS** — 6 `.gz` sorted-equivalent vs Perl (SPEC §8.3 `.gz` contract); M-bias.txt + splitting_report.txt strictly byte-identical ✅
- **Dual independent code-review: APPROVE** (no Critical/High). Documented non-blocking follow-ups: gzp's eager thread-pool footprint (`(threads+1)×open_files` ≈ 60 in Default mode), panic-on-drop footer-error (ENOSPC-only), SE-only test.

## Notes
- `.gz` raw bytes differ from the old flate2 output (different framing + `deflate_rust` skips the cross-block dictionary), but no test hashes raw `.gz` and the smoke compares `zcat|sort|md5`.
- The gzip footer is written on Drop (gzp's `Drop`→`finish()`), matching the prior flate2 lifetime the empty-sweep relies on.

Part of #884 (R2).

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -110,7 +110,7 @@ dependencies = [
  "rustc-hash",
  "smallvec",
  "tempfile",
- "thiserror",
+ "thiserror 2.0.0",
 ]
 
 [[package]]
@@ -123,12 +123,13 @@ dependencies = [
  "clap",
  "crossbeam-channel",
  "flate2",
+ "gzp",
  "mimalloc",
  "noodles-core 0.20.0",
  "noodles-sam",
  "predicates",
  "tempfile",
- "thiserror",
+ "thiserror 2.0.0",
  "which",
 ]
 
@@ -147,7 +148,7 @@ dependencies = [
  "proptest",
  "smallvec",
  "tempfile",
- "thiserror",
+ "thiserror 2.0.0",
 ]
 
 [[package]]
@@ -199,6 +200,12 @@ dependencies = [
  "regex-automata",
  "serde",
 ]
+
+[[package]]
+name = "bumpalo"
+version = "3.20.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72f5acc6cb2ba439de613abc23857ec3d78374d8ed5ac84e9d11336e87da8649"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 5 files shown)


FelixKrueger:
The three non-blocking follow-ups noted above (gzp eager thread footprint, panic-on-drop footer-error, SE-only gzip test) are tracked in #889 — deferred per review, not blocking this PR.

Issue #707, opened 2024-10-17
Speed up merge_individual_BAM_files


MilosCRF:
Hi Felix,

I’m currently running Bismark on AWS using 192 CPUs. Everything is running at lightning speed except for merging the temporary BAM files. It seems that the merging process is not utilizing multiple cores with the current `bismark` settings.

Do you know of any way to speed up this step? Perhaps adding `-@ $num_threads` to the `samtools `command that is likely handling the merging?

FelixKrueger:
`samtools cat` does have a `-@` flag, but I am not sure if this would at any point break the order of Read1/Read2 following each other directly (which would break downstream processed). Do you have possibility to find the command that merges the BAM files and try out adding increasing values of `-@`?

MilosCRF:
Thank you, Felix.
Yes, I can definitely test it. However, I couldn't find `samtools cat` in Bismark's main script. Isn't the merge accomplished by samtools view at line 1427?

`open (OUT,"| $samtools_path view -bSh 2>/dev/null - > ${output_dir}${merged_name}") or die "Failed to write to $merged_name: $!\n";`

FelixKrueger:
I think you are right, I must have confused this with something else (e.g. deduplication of multiple files).

Can you try to try a different number of threads? E.g. 2, 4, 8? We only have to be mindful how this would affect resource allocation for e.g. nf-core workflows, but this would be a downstream problem.

MilosCRF:
I apologize for the delay.

I conducted a test by altering the number of CPUs for `samtools view`, and the results are as follows:

**Time Efficiency:** The time required to merge temporary files decreased significantly (1 CPU = ~10 minutes vs. 4 CPUs = ~2.5 minutes). This improvement is substantial as the entire Bismark run originally took about 30 minutes for each of my samples (approximately 50M paired reads).
**CPU Utilization:** Utilizing more than 4 CPUs did not yield any further improvements. I am unsure why this is the case.  I had 12 temporary BAMs to merge as set by --parallel 12. Despite setting more than 4 CPUs, `samtools view` utilized only up to 4 CPUs for merging.
**Data Integrity:** The methylation profiles remained unaffected by the increase in the number of CPUs used for merging temporary BAMs.

A brief report from the test is attached.
[Bismark_mergeBAMs_multithreading.zip](https://github.com/user-attachments/files/17629523/Bismark_mergeBAMs_multithreading.zip)

FelixKrueger:
[...]

Issue #364, opened 2020-08-07
Add multicore option to deduplicate_bismark


insectopalo:
There are several call to samtools in the deduplicate_bismark script that would benefit from the -@ option that samtools has.

FelixKrueger:
Hi @insectopalo 

This is true and could almost certainly be used for single-end alignments, but I am not sure whether it would work for paired-end alignments, as we need to guarantee that the script processes two consecutive lines at a time. I think this could would require some testing once I am back from holidays.

FelixKrueger:
Sorry for the long delay, I have now run a few experiments regarding this issues, by adding additional threads to `samtools view`, both to the reading and writing file handles. I monitored the CPU usage via `top`,  and also measured the total time it took to process ~20 million single-end reads (on our head node...).

I am afraid the CPU usage for both reading and writing seem to be mostly well below 100%, irrespective of wether I used 1, 2 or 4 cores. Therefore, it doesn't seem like allowing Samtools to use more CPU system resources will result in a noteworthy performance increase.

To be perfectly honest, I'd therefore be tempted to just leave it as it is, unless you disagree vehemently?

**1 core:**
3m5.778s
3m1.591s
3m5.924s

**2 cores:**
2m36.280s
2m35.216s
2m37.697s

**4 cores:**
2m44.165s
2m37.788s
2m31.628s

FelixKrueger:
I'll interpret this as 'it doesn't look like it really makes a big difference'. So I'll go ahead and close this issue for now.

PR #1004, opened 2026-06-21
perf: parallelise BAM/gzip I/O + speed up bismark per-read loops (bit-identical)

BenjaminDEMAILLE:
## Summary

Two independent layers of speed-ups for the Perl Bismark tools, both verified to leave **every output byte-identical** (alignments, deduplicated reads, methylation calls):

1. **I/O parallelism** — thread the BAM/CRAM (de)compression that was still single-threaded, and use `pigz` for gzip when available.
2. **Per-read hot loops in `bismark`** — replace per-character array/`substr` work with single C-level string operations.

## Part 1 — I/O parallelism (bit-identical)

- Pass `--threads` to the single-instance/ambiguous `samtools view` BAM/CRAM writes (previously only the multi-instance merge was threaded, since #707).
- Use `pigz` in place of `gzip` for temporary/auxiliary and `--gzip` output streams when it is on `PATH` (automatic fallback to `gzip`).
- `deduplicate_bismark --parallel` was hard-coded to `1` and inert; it now controls the samtools thread count (default still `1`, so behaviour is unchanged unless opted in, consistent with #364).

Isolated component speed-ups (~399 MB payload, 16-core, best of 3):

| Operation | before (1 thread) | after (4) | speed-up |
|---|---|---|---|
| BAM compression `samtools view -bSh` | 4.25 s | 1.13 s | 3.8× |
| gzip → pigz | 18.8 s | 4.98 s | 3.8× |

## Part 2 — `bismark` per-read loops (bit-identical)

Profiling a single-instance run (NYTProf, 100k reads) showed ~90% of Bismark's own Perl time in four per-read subs. Each was rewritten to the same result with cheaper primitives:

- `hemming_dist` → `($a ^ $b) =~ tr/\0//c` instead of splitting both strings into arrays (~30× in isolation).
- `methylation_call` → start from an all-`.` string and visit only genomic `C`/`G` positions via `index()`, instead of splitting and branching at every base (~3×).
- `make_mismatch_string` (MD:Z) → locate mismatches with a single `($a ^ $b) =~ /[^\0]/g` scan instead of a per-base `substr` loop (~10×).
- `check_results_single_end` / `_paired_end` → split each aligner output line once instead of twice.

Each is guarded to fall back to the original code path on any length mismatch, and was fuzz-tested for identity (200k–300k randomized reads per sub) in addition to the pipeline check.

**End-to-end effect** — `bismark` alignment of 1M single-end reads, single instance:

| | wall-clock |
|---|---|
| before | 61.7 s |
| after | **36.7 s** |

≈ **40% faster** alignment (1.7×), entirely from the per-read loop rewrites.

## Correctness verification

Full SE + PE pipeline (align → deduplicate → methylation-extract, `--gzip --CX --bed`) on the bundled `test_files/` data, run with the current `master` and with this branch, comparing **decompressed, sorted** output:

- all BAMs identical (`samtools view | sort`)
- all gzip outputs identical (context files, bedGraph, coverage)
- report statistics identical (only the wall-clock "completed in" line differs)

## Notes / scope

- Profiling showed the per-read Perl bottleneck is concentrated in `bismark`; `deduplicate_bismark` (~2.4 µs/record) and `bismark_methylation_extractor` (sort/IO-bound, with an existing linear-CIGAR fast path) are not per-read-Perl-bound to the same degree, so no risky rewrites were applied there.
- Docs updated under `docs/options/`.


BenjaminDEMAILLE:
Superseded by two focused PRs to keep the changes independent and reviewable:
- #1007 — I/O parallelism (samtools --threads + pigz), bit-identical
- #1006 — bismark per-read loop speed-ups (~40% faster alignment), bit-identical

Closing this combined PR in favour of those.

PR #1006, opened 2026-06-21
perf(bismark): speed up the four hottest per-read loops (~40% faster alignment, bit-identical)

BenjaminDEMAILLE:
## Summary

Speeds up the `bismark` aligner by rewriting the four hottest per-read Perl subroutines, with **byte-identical output** (alignments and methylation calls are unchanged). This is purely CPU/algorithmic and is independent of the I/O-threading work in #1004.

## Motivation

Profiling a single-instance run (Devel::NYTProf, 100k reads) showed ~90% of Bismark's own Perl time (i.e. excluding the aligner subprocess) concentrated in four per-read subs. On a small genome where Bowtie2 is fast, this Perl overhead is a large share of wall-clock; it scales with read count, so it matters most for WGBS.

## Changes (each bit-identical, with a fallback)

- **`hemming_dist`** — count mismatches with `($a ^ $b) =~ tr/\0//c` instead of splitting both strings into per-character arrays. ~30× faster in isolation. Falls back to the original loop if the two strings ever differ in length.
- **`methylation_call`** — start from an all-`.` call string and visit only the genomic `C` (CT read) / `G` (GA read) positions via `index()`, instead of splitting both strings and branching at every base. Identical call string and all eight context counters. ~3× faster in isolation.
- **`make_mismatch_string`** (MD:Z tag) — locate genomic mismatches in one C-level pass with `($a ^ $b) =~ /[^\0]/g` instead of a per-base `substr` loop; `X` padding handling and deletion processing unchanged. ~10× faster in isolation. Length-guarded fallback to the original loop.
- **`check_results_single_end` / `check_results_paired_end`** — split each Bowtie2/HISAT2 output line once and reuse the fields, instead of splitting the same line twice.

Each rewrite was fuzz-tested for identity against the original (200k–300k randomized reads per sub, including `N`/`X` padding and random mismatches) in addition to the full-pipeline check below.

## End-to-end effect

`bismark` alignment of 1,000,000 single-end reads, single instance, best of 3:

| | wall-clock |
|---|---|
| `master` | 61.7 s |
| this branch | **36.7 s** |

≈ **40% faster** alignment (1.7×). The same per-read savings apply per instance under `--parallel`.

## Correctness verification

Full SE + PE pipeline (align → deduplicate → methylation-extract) run with `master` and with this branch on the bundled `test_files/` data; all output BAMs are byte-identical after `samtools view | sort`. Methylation extraction and downstream files are unaffected because the SAM records (including the `XM`/`MD`/`NM` tags) are unchanged.

## Notes

- `deduplicate_bismark` (~2.4 µs/record) and `bismark_methylation_extractor` (sort/IO-bound, with an existing linear-CIGAR fast path) are not per-read-Perl-bound to the same degree, so no rewrites were applied there.

### DIFF
--- a/bismark
+++ b/bismark
@@ -9270,40 +9270,42 @@ sub make_mismatch_string{
 	my $ref_base;
 	my $actual_base;
 	
-	foreach my $pos ( 0..(length$actual_seq) - 1 ){
-
-		$actual_base = substr($actual_seq,$pos,1);
-		$ref_base    = substr($ref_seq,$pos,1);
-		# if ($verbose){ warn "reference: $ref_base\tseen base: $actual_base\n";}
-
-		if ( $actual_base eq $ref_base ){
-			++$prev_matching;
+	### Build the MD:Z string. The genomic mismatches (non-'X') are exactly the positions where the read and
+	### reference bytes differ, which we locate in a single C-level pass via the XOR of the two strings rather than
+	### an explicit per-base substr loop. 'X' positions (artificial padding for insertions/soft-clipping) are skipped,
+	### exactly as before. Produces a byte-identical MD tag (verified on the full pipeline and >300k fuzzed reads).
+	if (length($actual_seq) == length($ref_seq)){
+		my $xor = $actual_seq ^ $ref_seq;
+		my $prev = -1;
+		while ($xor =~ /[^\0]/g){
+			my $mpos = pos($xor) - 1;
+			$prev_matching += ($mpos - $prev - 1);   # matching bases in the gap since the last emitted mismatch
+			$ref_base = substr($ref_seq,$mpos,1);
+			unless ($ref_base eq 'X'){               # 'X' padding (insertion/soft-clipping) is ignored for the MD tag
+				$MD_tag .= $prev_matching;
+				$MD_tag .= $ref_base;
+				$prev_matching = 0;
+			}
+			$prev = $mpos;
 		}
-		else{
-			# If the mismatch is due to an insertion or soft-clipping we simply move on, else we print the previously matching bases
-			# as well as the mismatching genomic base
-			if ($ref_base eq 'X'){
-				if ($verbose){ warn "The genome base was an artificially padded '$ref_base' due to an insertion or soft-clipping in the read at this position. Just ignoring it for the MD tag\n";}
+		$prev_matching += (length($actual_seq) - 1 - $prev);   # trailing matches after the last mismatch
+	}
+	else{
+		### Fallback preserving the exact original behaviour should the two strings ever differ in length.
+		foreach my $pos ( 0..(length$actual_seq) - 1 ){
+			$actual_base = substr($actual_seq,$pos,1);
+			$ref_base    = substr($ref_seq,$pos,1);
+			if ( $actual_base eq $ref_base ){
+				++$prev_matching;
 			}
 			else{
-				if ($verbose){ warn "previous matching bases: $prev_matching\n";}
-
-				### There is a mismatch between the sequence and the genome. First we need to write out how may bases matched until now
-				if ($prev_matching == 0){
-					if ($verbose){ warn "Got a mismatch either at the very start or next to another mismatch. Need to add a padding 0 as well as the mismatch\n";}
-					if ($verbose){ warn "${prev_matching}$ref_base\n";}
+				unless ($ref_base eq 'X'){
 					$MD_tag .= $prev_matching;
 					$MD_tag .= $ref_base;
+					$prev_matching = 0;
 				}
-				else{
-					if ($verbose){ warn "${prev_matching}$ref_base\n";}
-					$MD_tag .= $prev_matching;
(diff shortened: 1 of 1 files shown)


BenjaminDEMAILLE:
@FelixKrueger keen to hear your view on whether this is something you would merge. It rewrites the four hottest per-read subroutines in `bismark` (`methylation_call`, `make_mismatch_string`, `hemming_dist`, and the line parsing in `check_results_*`) using cheaper string primitives, for **byte-identical output**.

How it was checked, since the methylation calling is obviously sensitive:
- each rewritten sub was fuzz-tested against the original implementation (200k–300k randomised reads per sub, including `N`/`X` padding, conversions and random mismatches) — identical call strings, MD tags and all context counters;
- the full SE+PE pipeline output is byte-identical to `master` (`samtools view | sort`);
- each rewrite keeps a fallback to the original code path if the read/reference strings ever differ in length.

End-to-end this was ~40% faster single-instance alignment on a 1M-read benchmark (61.7s → 36.7s); the same per-read saving applies per instance under `--parallel`. The profiling pointed at these subs specifically (Devel::NYTProf), so I deliberately left `deduplicate_bismark` and `bismark_methylation_extractor` alone — they were not per-read-Perl-bound in the same way.

Draft release note in your CHANGELOG style for `v0.25.1dev`:

```
### Bismark
- sped up the per-read methylation calling, MD:Z generation, edit-distance and SAM-line parsing routines; output is byte-identical (verified by per-sub fuzzing and a full SE+PE pipeline diff). Roughly 40% faster single-instance alignment in a 1M-read benchmark
```

I realise byte-identical claims on the methylation caller need a careful eye, so very happy to add more test vectors or restructure if you would prefer a more conservative subset. Thanks!

FelixKrueger:
Hi @BenjaminDEMAILLE 

Many thanks for your PRs; this would - at least in theory - have been a welcome contribution. However, over the past month or so I have reimplemented the [entire Bismark suite in Rust](https://felixkrueger.github.io/Bismark/rust/overview/), and do not intend to maintain or improve the Perl version beyond v0.25.1 going forward (I may apply potential bug fixes though).

The latest version as of today is `v2.0.0_beta.11` which you can install and try out as [described here](https://felixkrueger.github.io/Bismark/installation/#install-from-source-with-cargo). 

For a rough comparison of performance and new/additional features please take a look at this [Bismark Rust suite benchmarks](https://felixkrueger.github.io/Bismark/rust/benchmarks/). Note that this is still a very active area of development, so expect frequent changes and additions.

I really appreciate your motivation to make Bismark a better (and faster) tool, but it so happens that I think this is already the case in the Rust version. I have paid particular attention to optimizing de-compression and compression questions with optimal parallelism (in addition to adding parallel worker threads where it makes sense). Please take a look and let me know if you have any questions.

Cross-linking #1007.

BenjaminDEMAILLE:
> Hi @BenjaminDEMAILLE
> 
> Many thanks for your PRs; this would - at least in theory - have been a welcome contribution. However, over the past month or so I have reimplemented the [entire Bismark suite in Rust](https://felixkrueger.github.io/Bismark/rust/overview/), and do not intend to maintain or improve the Perl version beyond v0.25.1 going forward (I may apply potential bug fixes though).
> 
> The latest version as of today is `v2.0.0_beta.11` which you can install and try out as [described here](https://felixkrueger.github.io/Bismark/installation/#install-from-source-with-cargo).
> 
> For a rough comparison of performance and new/additional features please take a look at this [Bismark Rust suite benchmarks](https://felixkrueger.github.io/Bismark/rust/benchmarks/). Note that this is still a very active area of development, so expect frequent changes and additions.
> 
> I really appreciate your motivation to make Bismark a better (and faster) tool, but it so happens that I think this is already the case in the Rust version. I have paid particular attention to optimizing de-compression and compression questions with optimal parallelism (in addition to adding parallel worker threads where it makes sense). Please take a look and let me know if you have any questions.
> 
> Cross-linking #1007.

Thanks !!
Will check ! 
Do you think Rust version is ready for production ?

FelixKrueger:
I am successfully using it for production stuff at work, using this fork of the nf-core/methylseq workflow, and adding -profile bismark-rust-profile https://github.com/FelixKrueger/methylseq/tree/bismark-rust-profile. Works like a charm!

FelixKrueger:
I am still doing some benchmarks, but will try to make the transition for the full v2.0.0 within the next weeks

BenjaminDEMAILLE:
> I am still doing some benchmarks, but will try to make the transition for the full v2.0.0 within the next weeks

Oh very cool !!! 
I have a M4 Max Mac so maybe I can try to do a PR for Apple Silicon optimisation at this moment

FelixKrueger:
In case you and your AI friend are keen to do something, please base everything off of the iron-chancellor branch.

BenjaminDEMAILLE:
> In case you and your AI friend are keen to do something, please base everything off of the iron-chancellor branch.

Will do a  new specific branch so

PR #1007, opened 2026-06-21
perf: parallelise BAM/gzip I/O (samtools --threads + pigz), bit-identical

BenjaminDEMAILLE:
## Summary

Parallelises the **I/O compression** steps of the Perl Bismark tools, which were largely single-threaded. Output content is **byte-identical** (alignments, deduplicated reads and methylation calls are unchanged). This is independent of the per-read loop speed-ups in #1006.

Applied to `bismark`, `deduplicate_bismark` and `bismark_methylation_extractor`:

1. **Threaded BAM/CRAM (de)compression** — pass `--threads` to the `samtools view` calls that were still single-threaded. In `bismark` only the multi-instance *merge* was threaded (since #707); the single-instance BAM/CRAM writes, the ambiguous-BAM write and several reads were not.
2. **`pigz` for gzip** — when `pigz` is on `PATH` it replaces plain `gzip` for temporary/auxiliary and `--gzip` output streams, with automatic fallback to `gzip`.

`deduplicate_bismark --parallel` was hard-coded to `1` and inert; it now controls the samtools thread count (default still `1`, so behaviour is unchanged unless opted in, consistent with #364).

## Benchmark (isolated components)

The end-to-end win shows where compression is a meaningful fraction (large BAM writes, `--gzip`/unmapped output); the per-read Perl loop otherwise dominates `deduplicate_bismark`/`bismark_methylation_extractor` (see #364). Isolated, on a ~399 MB payload (16-core, best of 3):

| Operation | before (1 thread) | after (4) | speed-up |
|---|---|---|---|
| BAM compression `samtools view -bSh` | 4.25 s | 1.13 s | 3.8× |
| gzip → pigz | 18.8 s | 4.98 s | 3.8× |

## Correctness verification

Full SE + PE pipeline (align → deduplicate → methylation-extract, `--gzip --CX --bed`) on the bundled `test_files/` data, run with `master` and with this branch, comparing **decompressed, sorted** output:

- all BAMs identical (`samtools view | sort`)
- all gzip outputs identical (context files, bedGraph, coverage)
- report statistics identical (only the wall-clock "completed in" line differs)

A regression caught during verification (an uninitialised thread count in the SE/PE header-detection read, which runs inside `process_commandline()`) is fixed in the last commit; output is bit-identical after the fix.

## Notes

- Default thread count for `bismark` is `4`, mirroring the merge step's value since #707. `bismark_methylation_extractor` uses a small fixed value (2, not scaled by `--multicore`) so total threads stay bounded given its fork-per-child design.
- Docs updated under `docs/options/`.


BenjaminDEMAILLE:
@FelixKrueger would you be open to merging this? It is purely I/O parallelism with **byte-identical output** (verified on the full SE+PE align → dedup → extract pipeline against `master`), so it should be safe to take independently of #1006.

A couple of points I am happy to adjust to your preference:
- the default `samtools --threads` value is `4` in `bismark`, mirroring the merge step you set in #707;
- `pigz` is auto-detected and used only when present (silent fallback to `gzip`) — if you would rather gate it behind an explicit `--pigz` flag, that is an easy change;
- `deduplicate_bismark --parallel` is unblocked but still defaults to `1`, so behaviour is unchanged unless opted in (consistent with #364).

If you would like a release note, here is a draft entry in the CHANGELOG style for `v0.25.1dev`:

```
### Bismark
- single-instance BAM/CRAM writes now use `samtools view --threads` (previously only the BAM merge was threaded, see #707); `pigz` is used for gzip compression of temporary/auxiliary files when it is available in the PATH (silent fallback to `gzip`)

### deduplicate_bismark
- `--parallel` now controls the number of `samtools` (de)compression threads (it was previously fixed to 1); `pigz` is used for the gzip fallback when available

### bismark_methylation_extractor
- BAM/CRAM input reads are now threaded, and `pigz` is used for `--gzip` output streams when available
```

Happy to tweak anything. Thanks for taking a look!

FelixKrueger:
HI @BenjaminDEMAILLE,

You beat me to it, but it is a Sunday after all.... I will reply to the first PR from earlier today.

PR #1013, opened 2026-06-22
perf(aligner): mimalloc fixes `--multicore` allocator-contention anti-scaling (3.1x), + byte-identical parse/output cleanups

BenjaminDEMAILLE:
# perf(aligner): mimalloc fixes `--multicore` allocator-contention anti-scaling (3.1x), + byte-identical parse/output cleanups

## What

Performance work on `bismark-aligner`, profiled on Apple Silicon (M4 Max),
**byte-identical to Perl v0.25.1** on the faithful Bowtie 2 path. Everything here
is output-neutral (no new flag, no concordance gate).

### The headline: mimalloc

`bismark-aligner` was the only hot crate without a multithreaded allocator (4
sibling crates already use mimalloc). Under `--multicore`, the worker threads
contended on the system allocator's arena locks, an anti-scaling pathology:

| RRBS-10M, GRCm39, M4 Max | wall |
|---|---|
| `-p 6` (1 Rust pipeline) | 1161 s |
| `--multicore 4`, system allocator | **5025 s** (4.3x SLOWER than `-p 6`) |
| `--multicore 4`, mimalloc | **1603 s** (3.13x faster, contention gone) |

Under `-p` the aligner is ~90% blocked on bowtie2, so the Rust allocator is off
the critical path. A first un-interleaved median-of-3 *looked* ~6% slower under
`-p 6`, but a **cooled, interleaved attribution run** (GATE_04: base vs
mimalloc-only vs full-stack, shared thermal state, 4-min cooldown per run) shows
that was a **thermal artifact**: the full stack is **statistically
indistinguishable** from baseline under `-p` (−1.3% median, inside a 14–19%
per-binary thermal spread). So there is **no `-p` trade-off** — mimalloc has no
measurable single-thread cost, just the large `--multicore` win. Absolute
wall-times should still be re-confirmed on the **Linux x86_64 benchmark host**
(no laptop throttling). The change is **byte-identical** and **bit-reproducible**
(`SOURCE_DATE_EPOCH`, same as the 4 sibling mimalloc crates). See `BENCHMARKS.md`
+ `GATE_04_p_attribution.md` for the full tables.

### Minor byte-identical cleanups (honest framing)

- **`SamRecord::parse` byte-level field split**: replaced `str::split('\t')`
  (`CharSearcher`, per-char decode) with a byte scan over the already-validated
  `&str`. **2.0x faster at the function level** (criterion: 278 ns -> 136 ns), but
  **within end-to-end noise** because bowtie2 dominates. Zero ripple (signature
  unchanged), byte-identical.
- **`build_pe_mate`**: dropped two per-mate intermediate allocations (an `md`
  scratch `Vec` the SE path already avoided, and an `md_value` `String` copy).
  Byte-identical; end-to-end within noise.

Profiling **ruled out** vectorizing the C->T/G->A transform (it is ~0.2% of wall;
GATE_00) — a profile-gated non-action.

## Why it is framed honestly

Under `-p` (the common single-pipeline usage), the aligner is **bowtie2-bound**:
~90% of `bismark_rs` wall is blocked in `read()` waiting on the external bowtie2,
~10% is Rust CPU (GATE_00 profile). So **no `-p` speedup is claimed** from
Rust-side work. The win is the `--multicore` allocator-contention fix, which is
exactly the regime the suite's scaling benchmarks use. The parse/output cleanups
are free, byte-identical, and faster at the function level, so they stay, but
they are not over-claimed as an end-to-end speedup.

## Validation

- **Byte-identity (Rust-vs-Rust)**: a new `#[ignore]`d oracle
  (`tests/byte_identity_real_data.rs`) compares the full BAM record stream
  (12,558,088 records) + report against a golden; verified identical across
  mimalloc + parse + build_pe_mate, and across `-p` vs `--multicore` (also proves
  parallel-invariance).
- **Byte-identity (Rust-vs-Perl)**: the full stack (mimalloc + parse +
  build_pe_mate) vs Perl Bismark v0.25.1 on RRBS / GRCm39: **246,856 records
  identical, report identical** (sorted `samtools view` + report diff, the
  `run_gate.sh` methodology run inline since that script trips macOS bash 3.2's
  `set -u` + empty-array bug).
- **Reproducibility**: `bismark_rs` built twice (clean) under `SOURCE_DATE_EPOCH`
  is bit-identical.
- **CI**: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`,
  `cargo test --workspace` all green.

## Notes for review

- The work is profiled on M4 Max only; the mimalloc win should re-measure on the
  Linux x86_64 benchmark host (it is universal, not ARM-specific — the extractor's
  mimalloc win was on Linux).
- Adds a `profiling` cargo profile (symbols for samply; never built by CI), a few
  `just` perf recipes, a `criterion` dev-dependency + one bench, and a documented
  **opt-in** `just build-native` (`RUSTFLAGS=-C target-cpu=native`; never the
  committed/CI default, to keep release binaries portable/reproducible).
- Provenance + per-change measurements live in `plans/06222026_aligner-apple-silicon/`.

### DIFF
--- a/plans/06222026_aligner-apple-silicon/BENCHMARKS.md
+++ b/plans/06222026_aligner-apple-silicon/BENCHMARKS.md
@@ -31,22 +31,29 @@ RSS stays ~3 GB, so it is lock-bound, not memory-bound). mimalloc removes the
 contention and `--multicore` becomes usable. This is the regime the suite's
 scaling benchmarks use.
 
-## `-p 6` (single Rust pipeline, 2 bowtie2 × 6 threads) — bowtie2-bound
-
-| `-p 6` (median of 3) | wall | peak RSS |
-|---|---|---|
-| BASE | 1145 s (1127–1158) | 3.0 GB |
-| AFTER | 1217 s (1204–1239) | 2.9 GB |
+## `-p 6` (single Rust pipeline, 2 bowtie2 × 6 threads) — bowtie2-bound, within noise
 
 Under `-p`, the aligner is **~90 % blocked waiting on the external bowtie2**
 (`sample` profile: 73,702 `read`-syscall samples vs ~7,000 Rust-CPU samples), so
-the allocator/parse work is off the critical path. Here AFTER measured **~6 %
-slower** than BASE — most likely **mimalloc's single-thread overhead** (it trades
-a little uncontended-allocation speed for large multithreaded scaling) exceeding
-the parse/build_pe_mate savings. Attribution (mimalloc-only vs the full stack)
-needs a cooled, controlled run; flagged honestly rather than hidden. Net: mimalloc
-is a clear win for `--multicore` users and a small cost for pure single-pipeline
-`-p` users — the maintainer may want to weigh gating it on `--multicore`.
+the allocator/parse work is off the critical path. A first un-interleaved
+median-of-3 measured AFTER **~6 % slower** here, but a **cooled, interleaved
+attribution run** (GATE_04: 3 distinct binaries, shared thermal state, 4-min
+cooldown per run) showed that "−6 %" was a **thermal artifact** — the difference
+vanishes into noise:
+
+| `-p 6`, interleaved (GATE_04) | median | mean | per-binary spread |
+|---|---|---|---|
+| BASE (system allocator) | 1242 s | 1240 s | 3.4 % |
+| MIMA (mimalloc only) | 1270 s | 1226 s | 14.0 % |
+| AFTER (full PR stack) | 1226 s | 1223 s | 19.2 % |
+
+Deltas vs BASE: AFTER **−1.3 % median / −1.4 % mean**; MIMA **+2.3 % median /
+−1.1 % mean** (sign flips by statistic). The per-binary thermal spread (14–19 %)
+**dwarfs** every inter-binary difference (±2 %), and AFTER's range
+[1114, 1328] fully envelops BASE's [1218, 1259]. The three are **statistically
+indistinguishable**. Net: mimalloc has **no measurable `-p` cost** (its
+single-thread overhead is below this laptop's noise floor) — so there is **no
+trade-off to weigh** and **no gate needed**; it is a clean `--multicore` win.
 
 ## Function-level: the parse byte-split (criterion, thermal-independent)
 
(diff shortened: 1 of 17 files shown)


FelixKrueger:
Thanks for this, @BenjaminDEMAILLE — and especially for steering the effort onto the Rust aligner; this is exactly the kind of contribution we're hoping for now that the Perl side is frozen.

I gave the changes a careful review (two independent passes). **Verdict: approve on the merits** — every change is byte-identical / output-neutral, which is the bar that matters for the faithful port.

**Byte-identity — verified, not just trusted:**
- **`SamRecord::parse` byte-scan**: equivalent to the old `trim_end_matches(['\n','\r'])` + `split('\t')` across every edge case (empty / leading / trailing-tab fields, lone `\r`, `\r\n`, mid-line `\r`, no trailing newline, empty → `[""]`); UTF-8-safe since `\t`/`\n`/`\r` are all `< 0x80`, so slices never land mid-codepoint.
- **`build_pe_mate`**: the new conditionally-owned `md` borrow matches the old nested guards exactly (`strand == b'-' && cigar.contains('D')`), same single revcomp; and the `md_value` `String` → `&[u8]` change is byte-neutral — it flows into `BString::from(...)`, whose `From<&[u8]>` and `From<String>` produce identical bytes, so the `MD:Z` tag is unchanged.
- **mimalloc `#[global_allocator]`**: binary-only, pinned identically to the four sibling crates.

The honest framing is appreciated — explicitly *not* claiming a `-p` speedup, and showing the apparent −6% was a thermal artifact (GATE_04), is exactly right: under `-p` the aligner is bowtie2-bound, so Rust-side allocation is off the critical path.

**Before merge (process, not code):**
1. **CI hasn't run on the branch yet** — could you push so `fmt`/`clippy`/`test` plus the `perl-oracle` byte-identity job execute on the runner? Everything's been validated locally (M4 Max), but the real-runner gate is the merge bar here.
2. **Bench flag:** `criterion = { default-features = false }` disables `cargo_bench_support`, so the *committed* bench won't actually measure — the 277 → 136 ns figures can't come from the manifest as committed. Worth enabling (dev-only) so the number is reproducible by others.
3. Minor: the real-data oracle is `#[ignore]`d, so CI won't guard the byte-identity contract for this crate (consistent with the sibling crates — just noting it); a small unit test pinning the `parse` edge cases would lock them in; and the oracle's hardcoded default base path could take an env default.

**On the headline:** the mimalloc `--multicore` win (≈3.1×, killing the allocator-contention anti-scaling) is the real prize, and it's well-attributed. I'll re-confirm it on our Linux x86_64 benchmark host — it should be universal (the extractor's mimalloc win was on Linux), and it gives us a reason to refresh the `-p` vs `--multicore` guidance, since that anti-scaling was part of what made `--multicore` look strictly dominated.

Thanks again — clean, well-documented work.

BenjaminDEMAILLE:
Thanks for the careful review, @FelixKrueger. Pushed 4 commits addressing all three points:

- **Bench flag** (`8599bb4`): enabled `cargo_bench_support` (kept the heavy plotters/rayon HTML machinery off). `cargo bench -p bismark-aligner --bench parse_bench` now actually measures — `byte_scan` 129.7 ns vs `char_searcher` 248.8 ns = **1.92x**. Good catch: I reconciled the documented figures to these reproducible numbers, since the earlier 277 -> 136 ns was an ad-hoc run the committed manifest couldn't reproduce.
- **Parse edge cases** (`19aa627`): extracted the byte-scan trim+split into a `trim_and_split` seam (pure move, byte-identical) and added a **non-ignored** CI unit test (`trim_and_split_matches_char_based_reference`) asserting equivalence to `line.trim_end_matches(['\n','\r']).split('\t')` across exactly the cases you listed: empty / leading / trailing-tab fields, lone `\r`, `\r\n`, mid-line `\r`, no trailing newline, empty -> `[""]`. So CI now guards the contract even though the real-data oracle stays `#[ignore]`d.
- **Oracle root** (`b47b634`): added a shared `BISMARK_ALIGNER_REAL_DATA_DIR` env fallback so the oracle isn't pinned to my host layout.

Local gates green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, full `cargo test -p bismark-aligner` (523 tests). Byte-identity unchanged. Ready for the runner whenever you approve the workflow — and thanks for offering to re-confirm the mimalloc `--multicore` win on the Linux x86_64 host.

PR #1014, opened 2026-06-23
feat(aligner): paired-end minimap2 (concordance-gated)

BenjaminDEMAILLE:
## Summary

Wires **paired-end minimap2** in `bismark-aligner` (previously hard-rejected / deferred), as a **concordance-gated, NOT-byte-identical** path — the same status as `--rammap` / `--combined_index`. Enabled via `--minimap2 -1 R1 -2 R2 <genome>` with a never-silent notice. The faithful Bowtie 2 / HISAT2 PE machinery (merge / scoring / MAPQ / `XM`-`XR`-`XG` / BAM output) is reused unchanged and stays byte-frozen.

There is no Perl oracle (the Perl PE minimap2 path is unfinished WIP — `# TODO`, `warn`+`sleep(1)` — and its report mislabels minimap2 as "HISAT2"), so it is gated against the **byte-frozen Bowtie 2 PE backend** instead.

## The load-bearing finding (verified minimap2 2.31)

With `-x map-ont` (Bismark's preset), minimap2 reads the two query files **sequentially** and emits **all read 1s, then all read 2s** — NOT interleaved mate pairs (and aligns them as independent single-end reads: SE-style FLAG 0/16/4, no proper-pair bits). So the consecutive-line pairing inherited from Perl is broken on real multi-read data (it crashed — exactly why the Perl path was left unfinished). A real-data gate caught this; the fake-binary tests had masked it.

## Implementation (all minimap2-gated / byte-neutral for Bowtie 2 / HISAT2)

- **`Minimap2PairedStream`** (align.rs): drains each instance, keeps the primary line per mate (skips `0x900` secondary/supplementary, keeps unmapped FLAG 4), and **joins read 1 ↔ read 2 by read-ID**, presenting one pair per read in input order (the order the merge lockstep walks). `drive_merge_pe` is generic over `PairedSamStream`; `process_pe_chunk` dispatches minimap2 to it.
- **Concordance enforcement** (merge.rs, gated on `Aligner::Minimap2`): both mates mapped, same chromosome, **FR orientation**, fragment within `--minins`/`--maxins` (default UNBOUNDED — long-read-oriented). Non-concordant "pairs" are dropped as no-alignment.
- **QNAME normalisation**: minimap2 doesn't clip read-number suffixes, so `Minimap2PairedStream` normalises its QNAME to the single-suffix Bowtie 2 shape — `from_lines` + `--ambig_bam` stay byte-frozen for all backends. (An earlier `/1/1`-then-`/1` strip over-stripped Bowtie 2 mate-suffixed reads → reverted.)
- `config.rs`: reject only rammap PE. `lib.rs`: never-silent concordance notice. Report correctly labels `minimap2` (a deliberate divergence from Perl's broken HISAT2 mislabel).

## Concordance gate (`plans/06232026_minimap2-pe/GATE.md` + `gate_harness.py`)

Real `bismark_rs` release binary, real Bowtie 2 2.5.5 + minimap2 2.31, simulated directional PE:

| Metric | Result |
|---|---|
| short-read at ground-truth position | **1000 / 1000 (100%)** |
| long-read PE at ground-truth (minimap2 niche) | **300 / 300 (100%)** |
| position concordance vs Bowtie 2 PE oracle | **100.000%** |
| `XM` methylation-call concordance vs Bowtie 2 | **100.000%** |
| determinism / worker-invariance (`--multicore 1`==`4`) | ✅ byte-identical |

## Tests / CI parity

- Unit: read-ID pairing + normalisation, concordance (FR / insert / cross-chromosome / half-mapped), `build_pe_argv` positional shape.
- Integration: `make_fake_minimap2_pe` emits sequential blocks; `minimap2_pe_mapped_names_and_report` + a multi-pair `minimap2_pe_sequential_blocks_pair_by_id` guard (would fail under consecutive-line pairing).
- `cargo fmt --check` clean; `clippy --workspace --all-targets --release -D warnings` clean; `cargo test --workspace` = 1983 passed / 0 failed. (The one pre-existing `--release` failure `inprocess::tests::reconstruct_cigar_rejects_pre_clipped_core_in_debug` — a `#[should_panic]` on a `debug_assert!` compiled out in release — is byte-identical to the base branch and unrelated.)

## Scope / honesty

The gate is local and small-scale (it demonstrates correctness + concordance + caught the real bug); a **maintainer full-scale real-data gate** remains the final production sign-off, as for every other backend. Short-read WGBS PE is already byte-identical via Bowtie 2 / HISAT2; minimap2 PE's real value is **long-read bisulfite PE** (covered by the 100% ground-truth-recovery metric). Exact byte-identity to Perl is permanently out of scope (no oracle).


FelixKrueger:
Hi Benjamin, 

I can see you are busy adding paired-end mode for minimap2. I have specifically rejected this mode so far for two reasons: 
- building a paired-end path for the long-read aligners would be implementing machinery for data that essentially doesn't exist in this space (neither ONT nor PacBio make paired-end reads?) 
- for short read data, people should not use `minimap2` (or `rammap`) but resort to the Bowtie2/HISAT2 routes
What does your Claude say about this?

BenjaminDEMAILLE:
Closing this. Felix is right to reject paired-end minimap2, and the exercise confirms it:

- ONT and PacBio do not produce paired-end reads at the FASTQ level (single-molecule; CCS/duplex read both strands of one molecule but emit a single read). There is no standard paired-end long-read library, and none for methylation — long-read methylation is native (ONT 5mC / PacBio kinetics), not bisulfite. So PE minimap2 in Bismark serves essentially no real data.
- For short-read bisulfite PE, Bowtie 2 / HISAT2 are the right (byte-identical) routes, as intended.

The work did surface one useful finding that supports the decision: with `-x map-ont`, minimap2 reads the two query files sequentially (all R1s, then all R2s), so the Perl `paired_end_…_minimap2` consecutive-line pairing cannot work as written — that's the unfinished `# TODO`. "Finishing" it would mean replacing the whole pairing model for data that doesn't exist. The hard-reject (with the "use --bowtie2/--hisat2 for paired-end" message) is the correct behavior and stays.

Branch kept as a reference; not merging.

Issue #1025, opened 2026-06-25
Faster FastQ inputs + alternative input file formats


ewels:
See https://github.com/FelixKrueger/TrimGalore/issues/315 (sample applies to both tools).

FelixKrueger:
👀 . on it

FelixKrueger:
Thanks Phil — dug into all four. Quick summary of where I've landed for Bismark specifically (the reasoning is a bit Bismark-flavoured because of how it's built).

**The architecture caveat that drives everything:** the Rust `bismark` aligner isn't an aligner — it's a wrapper. It reads the raw reads, bisulfite-converts them (C→T, plus the G→A complement), writes *converted* temp FASTQ, and shells out to Bowtie2/HISAT2/minimap2, which re-read those converted files themselves. Alignment is ~74% of the wall time, and the read+convert step that sits in front of it is single-threaded. So on the faithful path, a faster FASTQ *reader* only speeds a small, non-dominant slice — and we still have to write standard FASTQ for the external aligner regardless.

With that lens:

- **uBAM input — yes, this is the one I want to do.** You're right that it's not supported as aligner input yet. It's the highest-value item here because it's a *format-coverage* win (independent of the parsing-speed question), it's a genuinely common raw-read container now (ONT/PacBio basecallers, archival), and Bismark already depends on `noodles-bam` for the post-alignment tools, so reading name/seq/qual to feed the convert step is low-effort. This is top of the list.

- **paraseq — great library, but its edge is parallel-only.** By the author's own benchmarks it *matches* needletail/seq_io single-threaded (needletail often wins), and the throughput advantage only shows up in multi-threaded record-set use. Our convert step is serial + feeds an external aligner, so there's little to gain on the current path. Where it *would* shine is the pure-in-Rust/in-process aligner path we're prototyping (no external aligner, no temp rewrite) — which is also the regime your linked scverse/rustar-aligner#95 lives in. So: parked against that, not the wrapper.

- **mim — same story.** It's a real Rust crate now (`mim-index`), but it needs a one-time per-file `.mim` index build (~20% over a single-thread parse) before the parallel decode pays off. A one-pass aligner can't amortise that, so it only helps if the index is reused — again, more relevant to the in-process path.

[...]

PR #1029, opened 2026-06-26
feat(aligner): BINSEQ (.vbq) read input (#1025 Phase 2)

FelixKrueger:
## Summary

Adds **Arc Institute BINSEQ `.vbq`** as a read input for `bismark_rs` — the **2nd backend** of the pluggable-input front-end (#1025), plugged into the Phase-1 `InputDecoder` / `resolve_alt_inputs` seam (uBAM was the 1st). A `.vbq` is decoded **in-process** via the `binseq` crate (0.9.2, MIT) to a temp FASTQ that byte-matches `bqtools decode`, then fed to the **byte-frozen** convert→align→merge core — so a `.vbq` run is **identical to running Bismark on the equivalent `bqtools decode` output**, exactly as a uBAM run ≡ `samtools fastq`. SE + PE; all library types; all backends.

Builds on `rust/binseq-input @ d0816e3` (the seam refactor, already merged-style on the branch). This PR = the BINSEQ backend (`2afbe04`) + review fixes (`5b1036a`) + docs (`576aedd`).

## Key design (per plan Rev 1, R1–R8)

- **R1** serial, file-order decode via the concrete `vbq::MmapReader` (`new_block`/`read_block_into`/`block.iter()`) — NOT the parallel `BinseqReader`, which reorders blocks (gate-fail) and independently reorders R1/R2 (mate mis-pair). Uses `decode_s(&mut buf)` (the serial path never calls `decode_all`, so `sseq()` would panic). Determinism asserted.
- **R2** D2 reject at the **file-header** level (`FileHeader.qual`/`.headers`): the crate masks per-record absence (`?`-fills quality via `DEFAULT_QUALITY_SCORE`, synthesizes numeric headers), so a per-record check would silently emit `?`-quality / index-named FASTQ. Quality-less or name-less VBQ is rejected, never silently filled.
- **R3** byte format pinned from **bqtools 0.5.7 source** (`write_fastq_parts`): `@<header>\n<seq>\n+\n<qual>\n`; header stored WITHOUT `@` (we prepend), quality VERBATIM raw ASCII (no phred offset, unlike uBAM). Both sides call the same `binseq` decode, so values match by construction — only the line formatting is ours.
- **R5** PE = one record carries both mates → primary→R1, extended→R2 (no collation). **R6** temp `<stem>.fastq` / `<stem>_1/2.fastq`. **R4** extension detect `.vbq`/`.cbq`/`.bq`, disjoint from FASTQ/uBAM.
- **R7** feature-gated behind default-OFF **`binseq-input`** (mirrors `rammap-inprocess`); `default-features=false`. Detection is **always-compiled**, so a default build rejects `.vbq` never-silently ("compiled without BINSEQ support"). Release Linux binaries + the container enable the feature (`release.yml`).

## Scope / deviations (all surfaced + user-approved)

1. **VBQ only (v1).** `.cbq`/`.bq` are detected then **rejected fail-loud**: BQ has no quality/names (D2); CBQ's serial reader isn't cleanly exposed by binseq 0.9.x (only `process_parallel`) — a documented fast-follow.
2. **Workspace dep bumps** required by `binseq`: `thiserror` `=2.0.0`→`=2.0.18` (12 crates; binseq needs `^2.0.17`) and `anyhow` `=1.0.86`→`=1.0.103` (4 crates; binseq's CBQ EliasFano path needs the `anyhow` feature to compile). Patch-level, **byte-output-neutral** (error text is stderr-only, not in the byte-identity gate).
3. **R7 footprint note:** `default-features=false` does NOT drop zstd (non-optional) and `anyhow` must stay on to compile; the feature gate is what keeps zstd off the default build.

## Tests / verification (local — all green)

- fmt clean; clippy `-D warnings` clean (feature-off, feature-on, workspace).
- feature-on aligner: 442 lib + **8** `binseq_transcode` integ (byte-exact SE/PE, file-level quality/header reject, multi-block, determinism, empty, PE classification) + 7 uBAM integ + 100 conformance.
- feature-off **workspace: 1211 tests** (the dep bumps regress nothing).
- New `binseq-input` CI job (mirrors `rammap-inprocess`) compiles + clippies + tests the gated path.
- Dual code-review **APPROVE** (0 Critical/High) + plan-manager **COMPLETE**.

## ⚠️ Pending pre-merge validation (merge-when-green)

The **oxy equivalence gate** has NOT run yet (oxy was unreachable). Before merge:
- `bismark_rs <reads.vbq>` records **==** `bismark_rs <bqtools decode reads.vbq>`, SE + PE × {directional, non-directional, pbat} (bqtools-encoded 10M subsample).
- Full **uBAM re-gate (R8)** — the shared seam was touched by Phase 1's generalization.

Hermetic tests + the from-source bqtools-format pin give strong local confidence; the oxy gate is the real-data byte-identity confirmation. **Do not merge until it passes.**

Plan: `plans/06252026_pluggable-inputs/phase2-binseq-input/PLAN.md` (Rev 1 + Implementation Notes).


FelixKrueger:
## ✅ Oxy equivalence gate PASSED — 12/12 byte-identical in-order

Ran on oxy (GRCh38, Bowtie 2 2.5.5, 1,000,000 reads subsampled from the directional 10M SE/PE sets), commit `576aedd`, `bqtools` 0.5.7.

| Library | BINSEQ SE | BINSEQ PE | uBAM SE | uBAM PE |
|---|---|---|---|---|
| directional      | ✅ 848,122 | ✅ 1,703,286 | ✅ 848,150 | ✅ 1,703,342 |
| non-directional  | ✅ 847,431 | ✅ 1,703,188 | ✅ 847,445 | ✅ 1,703,244 |
| pbat             | ✅ 4,644   | ✅ 3,182    | ✅ 4,631   | ✅ 3,182    |

**0 sorted-only, 0 FAIL — every cell is exact in-order byte-identity.**

- **BINSEQ (6 cells):** `bismark_rs <reads.vbq>` records == `bismark_rs <bqtools decode -T1 reads.vbq>` records. (`-T1` = deterministic file-order decode; default parallel `bqtools decode` reorders blocks, which is why the in-process serial reader — plan R1 — is the right design.)
- **uBAM R8 re-gate (6 cells):** `bismark_rs <ubam>` records == `bismark_rs <samtools fastq ubam>` records — confirms the shared `resolve_alt_inputs` seam refactor did **not** regress the shipped uBAM path.

Notes: pbat counts are small (the test reads are directional → few map under pbat's G→A strands), but each cell sees the **same** records on both sides, so equivalence holds; non-dir cells exercise all 4 strands (CTOT/CTOB populated), identical both sides.

Combined with CI green (incl. `binseq-input` job + `perl-oracle byte-identity`) + dual code-review APPROVE + plan-manager COMPLETE, this PR is **ready to merge**. Gate detail: `plans/06252026_pluggable-inputs/phase2-binseq-input/GATE.md`.

PR #1015, opened 2026-06-23
feat(aligner): Illumina 5-Base (5mC→T) support — SE + PE, concordance-gated (#787)

BenjaminDEMAILLE:
## What

Opt-in support for **Illumina 5-Base** data (issue #787) in `bismark-aligner`, **single-end and paired-end**, behind `--illumina_5base` (alias `--five_base`).

5-Base is the chemical **inverse** of bisulfite: the enzyme converts **5mC → T** and leaves unmethylated C intact (confirmed by Illumina's data sheet M-GL-03689: "direct conversion of 5-methylcytosine to thymine"), so library complexity is preserved and reads align to the **unconverted** genome with a standard aligner. This path aligns the **raw** reads (no C→T/G→A conversion), derives the strand from the SAM FLAG, reuses the **byte-frozen** genomic extraction + `single_end_sam_output`/`paired_end_sam_output`, and runs `methylation_call` at **inverted polarity** (a read `T` at a genomic C = methylated). The emitted BAM is standard Bismark-convention `XM`/`XR`/`XG`, consumed unchanged by the extractor / bedGraph / coverage2cytosine / report.

## Supported path

- **SE + PE** core 5-Base alignment + inverted methylation call (minimap2 `-x sr` against the genome FASTA by default; `--bowtie2`/`--hisat2` via `--five_base_index <basename>`, a user-built NORMAL unconverted index).
- `--multicore` honored as a thread count (single instance, scale with `-p`).

## Experimental / preview modes

Wired end-to-end and never-silent, but **not byte-identity- or per-site-CI-gated** — labeled experimental in `--help` + README:

- **`--five_base_deconvolution`** — two-strand variant-vs-5mC deconvolution (DRAGEN's rule): a CpG gone on BOTH strands is a C>T/G>A **variant**, excluded from methylation. Per-CpG report.
- **`--five_base_duplex`** — per-molecule duplex reconciliation. **PE keys on the fragment span (POS+mate+TLEN)** — the real workflow (5-Base is paired-end). **SE-duplex is a known limitation** (SE OT/OB reads cover opposite fragment ends, different spans → never pair; degenerate non-workflow).
- **`--five_base_consensus`** — collapse each duplex family to one consensus read. Known limitation: the consensus record is forward/OT-only, so its `XM` carries `+`-strand CpG calls only (the duplex report covers both strands).
- **`--five_base_umi_qname`** — duplex dual-UMI from the read NAME (`A+B`, partner swapped) — the real-data UMI form.

## Validation

[...]

PR #1034, opened 2026-06-27
feat(aligner): opt-in paraseq + mim FastQ readers for the convert step (#1025 Phase 3b/3c)

BenjaminDEMAILLE:
## What

Two opt-in fast/opportunistic FastQ readers for the bisulfite-convert step of #1025 ("Faster FastQ inputs + alternative input file formats"). Both are additive, opt-in, never-silent, feature-gated; the byte-frozen convert -> align -> merge core and its Perl v0.25.1 byte-identity are untouched on the default path.

> **CBQ dropped from this PR (was 73d8262).** Per Felix's review, CBQ decode hit a **blocking upstream defect**: binseq 0.9.2 (+ sucds 0.8.3, both latest) mis-decodes N-containing `.cbq` — silent `N`->`A` in release, panic in debug. Reproduced exactly (`NACGTACGT`->`AACGTACGT`, `NNNNNNNN`->`AAAAAAAA`). N support is CBQ's whole reason to exist over VBQ, there is no clean public API to detect Ns in binseq 0.9.2, and a silent `A` where a methylation caller expects a no-call `N` is unacceptable. CBQ is parked (the correct decode plumbing is preserved) pending an upstream binseq fix + an N round-trip re-verification, and will be re-proposed in its own PR. An upstream bug report against binseq/sucds is being filed.

### Phase 3c, paraseq fast FastQ parse (`--paraseq`)
Opt-in: parse the input FastQ with the `paraseq` reader in the bisulfite-convert step instead of the built-in `read_until` loop. Byte-identical to the default path; the win is faster raw-read parsing.
- Serial `fastq::Reader` (not the parallel `process_parallel`, which exposes no record-set index, so it cannot guarantee byte-identical file order; the maintainer's spike also found parallel convert value-negative on the wrapper path). Records are converted + written inline, capturing the parse win without the channel-handoff regression.
- Behind the default-OFF / release-ON `paraseq-convert` feature. Forced one workspace pin bump: `smallvec` =1.13.2 -> =1.15.2 (byte-output-neutral).
- A `paraseq_byte_identity` gate asserts serial == paraseq across SE C->T/G->A, PE R1/R2, --icpc/--skip/--upto, CRLF, and gzip output.

### Phase 3b, mim opportunistic gzip index (`--mim`)
Opt-in: when an input `.fq.gz` has a pre-built `<file>.mim` sidecar, decode it in parallel (record-aligned chunk ranges, one per worker) in the convert step; otherwise emit a note and fall back to the standard parser.
- Byte-identical AND worker-count-invariant: mim assigns CONTIGUOUS ascending chunk ranges per worker; each worker drives our own `paraseq::fastq::Reader` over mim's `GzipStreamReader` and reuses the shared `emit_converted`, and buffers are concatenated in worker order = file order. `--skip`/`--upto`/count stay exact via each chunk's `next_record_idx`. With `--gzip` the temp is decompression-identical (deflate framing may differ; the decompressed reads, alignment, and final output are byte-identical).
- Behind the default-OFF / release-ON `mim-input` feature. UNLIKE the others it requires **Rust >= 1.91** (`mim-index` is edition 2024), so it gets a dedicated 1.91 CI job; the default build + all other features stay 1.89-valid. Forced one workspace pin bump: `clap` =4.5.30 -> =4.5.61 (behavior + CLI-output neutral; the suite uses custom error strings, not clap auto-help). Mutually exclusive with `--paraseq`.
- A `mim_byte_identity` gate builds a real `.fq.gz` + `.mim` sidecar and asserts serial == mim at 1/2/4 workers across SE C->T/G->A, PE, --skip/--upto, and --gzip.

## Test plan
- `cargo test --workspace` (default): all crates green (validates the clap bump across all 12 binaries).
- `cargo test -p bismark-aligner --features paraseq-convert` (paraseq concordance), `--features mim-input` (mim concordance).
- `cargo clippy -p bismark-aligner --all-targets --release [--features ...] -- -D warnings` clean on default + both features.
- MSRV: default workspace builds on **1.89**; `mim-input` builds on **1.91**. New CI jobs cover paraseq-convert (1.89) and mim-input (1.91); release ships both features on `@stable`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,13 +110,15 @@ jobs:
           BISMARK_SUITE_VERSION: ${{ needs.check-release.outputs.version }}
         # Compile the `bismark` aligner with the in-process rammap backend ON so the shipped
         # binary's `--rammap_inprocess` opt-in is functional (default `--rammap`
-        # stays subprocess), AND with `binseq-input` ON so the shipped binary can read
-        # BINSEQ `.vbq` files (#1025 Phase 2 — a default `cargo build` rejects `.vbq`
-        # fail-loud; release binaries decode it). `pkg/feature` syntax: both features
-        # live only on bismark-aligner; this still builds all 14 workspace members, and
-        # the other 11 binaries are byte-unaffected. The rammap-core git dep + the
-        # binseq crates.io dep are fetched at build (network available on the runner).
-        run: cargo build --release --locked --manifest-path rust/Cargo.toml --target ${{ matrix.target }} --features bismark-aligner/rammap-inprocess,bismark-aligner/binseq-input
+        # stays subprocess), with `binseq-input` ON so the shipped binary can read
+        # BINSEQ `.vbq`/`.cbq` files (#1025 Phase 2/3: a default `cargo build` rejects
+        # them fail-loud; release binaries decode them), AND with `paraseq-convert` ON so
+        # the shipped binary honours `--paraseq` (#1025 Phase 3c, opt-in fast FastQ parse,
+        # byte-identical; the default convert path is unchanged). `pkg/feature` syntax: all
+        # three features live only on bismark-aligner; this still builds all 14 workspace
+        # members, and the other 11 binaries are byte-unaffected. The rammap-core git dep +
+        # the binseq/paraseq crates.io deps are fetched at build (network available on the runner).
+        run: cargo build --release --locked --manifest-path rust/Cargo.toml --target ${{ matrix.target }} --features bismark-aligner/rammap-inprocess,bismark-aligner/binseq-input,bismark-aligner/paraseq-convert
       - name: Package the suite tarball
         id: package
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,13 +112,17 @@ jobs:
         # binary's `--rammap_inprocess` opt-in is functional (default `--rammap`
         # stays subprocess), with `binseq-input` ON so the shipped binary can read
         # BINSEQ `.vbq`/`.cbq` files (#1025 Phase 2/3: a default `cargo build` rejects
-        # them fail-loud; release binaries decode them), AND with `paraseq-convert` ON so
-        # the shipped binary honours `--paraseq` (#1025 Phase 3c, opt-in fast FastQ parse,
-        # byte-identical; the default convert path is unchanged). `pkg/feature` syntax: all
-        # three features live only on bismark-aligner; this still builds all 14 workspace
-        # members, and the other 11 binaries are byte-unaffected. The rammap-core git dep +
-        # the binseq/paraseq crates.io deps are fetched at build (network available on the runner).
-        run: cargo build --release --locked --manifest-path rust/Cargo.toml --target ${{ matrix.target }} --features bismark-aligner/rammap-inprocess,bismark-aligner/binseq-input,bismark-aligner/paraseq-convert
+        # them fail-loud; release binaries decode them), with `paraseq-convert` ON so the
+        # shipped binary honours `--paraseq` (#1025 Phase 3c, opt-in fast FastQ parse), AND
+        # with `mim-input` ON so it honours `--mim` (#1025 Phase 3b, opt-in opportunistic
+        # gzip-index parallel decode). Both are opt-in + byte-identical; the default convert
+        # path is unchanged. The release toolchain is `@stable` (>= 1.91), which satisfies
+        # mim-index's edition-2024 MSRV (the default/MSRV-1.89 CI builds never compile mim).
+        # `pkg/feature` syntax: all four features live only on bismark-aligner; this still
+        # builds all 14 workspace members, and the other 11 binaries are byte-unaffected. The
+        # rammap-core git dep + the binseq/paraseq/mim-index crates.io deps are fetched at
+        # build (network available on the runner).
+        run: cargo build --release --locked --manifest-path rust/Cargo.toml --target ${{ matrix.target }} --features bismark-aligner/rammap-inprocess,bismark-aligner/binseq-input,bismark-aligner/paraseq-convert,bismark-aligner/mim-input
       - name: Package the suite tarball
         id: package
         run: |
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -165,6 +165,14 @@ jobs:
         run: cargo build -p bismark-aligner --all-targets --features paraseq-convert
       - name: cargo clippy -p bismark-aligner --features paraseq-convert -D warnings
         run: cargo clippy -p bismark-aligner --all-targets --features paraseq-convert -- -D warnings
+      - name: Install minimap2 (5-Base #787 ground-truth gates)
+        # `cargo test -p bismark-aligner` runs the 5-Base real-aligner gates, which panic
+        # when $CI is set and minimap2 is missing (they must not pass vacuously). This
+        # feature job runs them too, so it needs minimap2 like the main `cargo test` job.
+        run: |
+          sudo apt-get update || sudo apt-get update
+          sudo apt-get install -y --no-install-recommends minimap2
+          minimap2 --version
       - name: cargo test -p bismark-aligner --features paraseq-convert
         run: cargo test -p bismark-aligner --features paraseq-convert
 
@@ -195,6 +203,14 @@ jobs:
         run: cargo build -p bismark-aligner --all-targets --features mim-input
       - name: cargo clippy -p bismark-aligner --features mim-input -D warnings
         run: cargo clippy -p bismark-aligner --all-targets --features mim-input -- -D warnings
+      - name: Install minimap2 (5-Base #787 ground-truth gates)
+        # `cargo test -p bismark-aligner` runs the 5-Base real-aligner gates, which panic
+        # when $CI is set and minimap2 is missing (they must not pass vacuously). This
+        # feature job runs them too, so it needs minimap2 like the main `cargo test` job.
+        run: |
+          sudo apt-get update || sudo apt-get update
+          sudo apt-get install -y --no-install-recommends minimap2
+          minimap2 --version
       - name: cargo test -p bismark-aligner --features mim-input
         run: cargo test -p bismark-aligner --features mim-input
 
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -139,6 +139,35 @@ jobs:
       - name: cargo test -p bismark-aligner --features binseq-input
         run: cargo test -p bismark-aligner --features binseq-input
 
+  paraseq-convert:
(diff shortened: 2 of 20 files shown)


FelixKrueger:
## CBQ (`73d8262`) review — careful implementation, but a **blocking** N-decode defect (upstream `binseq`, surfaced by the missing N test)

Reviewed the CBQ commit in depth (I did the Phase 2 VBQ backend + deferred CBQ at the time, so this is the piece I punted on). The decode *plumbing* is correct and careful — but building and running it locally turned up a blocking correctness bug in CBQ's headline feature.

**Strengths (the plumbing is right)**
- Correctly drives the CBQ **serial** reader (the part Phase 2 deferred as fragile): streaming `cbq::Reader` → `read_block` → `decompress_columns` → `iter_records`, with a correct `BlockRange` (`cumulative += bh.num_records`; the synthesized-index path is proven unreachable since headers are required). File-order + deterministic.
- Empty-`.cbq` handled by reading the fixed `FileHeader` directly rather than via `cbq::MmapReader` (which builds a footer index that fails on a zero-record file) → empty input → empty FASTQ, matching VBQ.
- Clean refactor (shared `common` module / DRY `write_fastq_record`), D2 reject via `has_qualities()`/`has_headers()`, PE primary→R1 / extended→R2 with the desync guard, **zero new deps / zero MSRV impact**, feature-gated with a never-silent feature-off reject.
- Built + ran the suite locally: `cargo test -p bismark-aligner --features binseq-input --test binseq_transcode` → **16 passed, 0 failed** (all 8 `*_cbq_*` cases); `cargo clippy --all-targets --features binseq-input -- -D warnings` clean.

**🔴 Blocking: N-containing `.cbq` does not decode correctly.**
The 8 tests all use ACGT-only sequences (inherited from the VBQ suite, where N was impossible under 2-bit). CBQ's entire reason to exist over VBQ is *native `N` support*. I added a throwaway test that writes N-containing reads via the canonical `binseq` CBQ writer and decodes them through this path:

| Build profile | Result on any N-containing `.cbq` |
|---|---|
| debug / `cargo test` | **hard panic** in `sucds-0.8.3 .../elias_fano/iter.rs:23` (`debug_assert_ne!(high_bits.num_ones(), 0)`) |
| **release** (the shipped `binseq-input` config) | **silent corruption — every `N` is emitted as `A`** (`NACGTACGT`→`AACGTACGT`, `NNNNNNNN`→`AAAAAAAA`; lead/mid/trail/consecutive all affected) |

Mechanism: CBQ stores bases 2-bit (N as placeholder `A`) plus an Elias-Fano index of N-positions, and `decompress_columns()` calls `backfill_npos()` to restore the `N`s. In `binseq 0.9.2` the deserialized npos Elias-Fano is degenerate (`len()` ≥ 1 but `high_bits.num_ones() == 0`), so the restore step asserts (debug) or yields nothing (release), leaving the placeholder `A`s. **This is inside `binseq`/`sucds`, not your code** — you call the documented decode correctly, and `bqtools decode` on the same file would corrupt identically (so the "byte-identical to `bqtools`" property holds — but it's byte-identical to a broken reference). For a methylation caller this is the worst failure class: an emitted `A` is treated as real signal where `N` is a no-call, and it silently violates the never-silent principle.

**No dependency bump fixes this:** `binseq 0.9.2` is already the latest release (2026-06-26) and it pins `sucds = "0.8.3"`, also the latest (2025-08-01). The defect is in current-latest `binseq`; the fix has to be a future upstream release.

**Recommendation**
- Don't merge CBQ as N-supporting until either (a) an upstream `binseq` fix ships and the N round-trip is re-verified, or (b) N-containing `.cbq` is rejected **fail-loud** with a clean error (detection likely needs a header/block-level npos check before `backfill_npos`, possibly a small upstream ask). Worth filing the bug upstream against `binseq`/`sucds`.
- Add an N-containing CBQ test regardless — in debug it *fails loudly* and would have caught this; it's the guard that keeps the feature honest once upstream is fixed.
- Still suggest **splitting CBQ into its own PR** from paraseq (3c) / mim (3b) — different value/risk profiles — but it can't land claiming N support in its current state.

Happy to share the exact throwaway repro (a `Format::Cbq` writer test with `N`s) if useful.

BenjaminDEMAILLE:
Thanks for the deep review, and for building + running the N case locally — you're exactly right, and I reproduced it both ways here:

- **debug:** panic at `sucds-0.8.3 .../elias_fano/iter.rs:23`
- **release** (the shipped config): silent `N`→`A` — `NACGTACGT`→`AACGTACGT`, `NNNNNNNN`→`AAAAAAAA`, `ACGTNACGT`→`ACGTAACGT`

My CBQ tests inherited the VBQ suite's ACGT-only fixtures, so they never touched CBQ's headline feature — that one's on me. And you're right that there's no clean public way to detect Ns in binseq 0.9.2 to reject them selectively (the npos fields are all `pub(crate)`), so a partial "N-free CBQ only" path isn't feasible either. A silent `A` where a methylation caller expects a no-call `N` is unacceptable, so:

- **CBQ is pulled from this PR** (rebased out; the decode plumbing is parked for re-proposal once upstream lands a fix + I add an N round-trip test). This PR is now **`--paraseq` + `--mim` only**, per your suggestion to split by value/risk.
- **Filed upstream:** ArcInstitute/binseq#94 (with the repro + the sucds degenerate-EliasFano mechanism).

paraseq (3c) and mim (3b) are unaffected — both byte-identical to the serial path (mim also worker-count-invariant), each with a serial-vs-feature concordance gate. Happy to split those into two PRs as well if you'd prefer.

BenjaminDEMAILLE:
Resolved + rebased onto current `rust/iron-chancellor`:

- **CBQ pulled** per your review (the N-decode defect is upstream `binseq`/`sucds`; filed ArcInstitute/binseq#94). This PR is now **`--paraseq` (3c) + `--mim` (3b)** only.
- **Rebased** onto the latest iron-chancellor (GA rename #1038 + version bump #1037 + 5-Base #1015): the PR was `CONFLICTING`, now `MERGEABLE`/`CLEAN`. Conflicts were all Cargo.toml version pins: kept iron-chancellor's GA `bismark-meta = "1.0.0"` pins and reconciled `clap` to `=4.5.61` workspace-wide (required transitively by `mim-index`'s `clap ^4.5.53`; byte-output-neutral, same precedent as the binseq `thiserror`/`anyhow` bumps). Dropped some stray `.DS_Store` that had crept into a commit.

Local: `fmt` + `clippy --release -D warnings` (default + `paraseq-convert,mim-input,binseq-input`) clean; `cargo test --workspace` + the feature-gated tests all pass on rustc 1.96. Both modes stay byte-identical to the serial convert path (mim also worker-count-invariant), each with a serial-vs-feature concordance gate.

Happy to split paraseq / mim into two PRs if you'd still prefer.

FelixKrueger:
Thanks for the careful work here, @BenjaminDEMAILLE — and for the fast CBQ turnaround (reproducing the N-decode defect both ways, filing [`ArcInstitute/binseq#94`](https://github.com/ArcInstitute/binseq/issues/94), and preserving the decode plumbing). The paraseq/mim implementations are genuinely clean: opt-in, never-silent, feature-gated, with serial-vs-feature concordance gates and worker-count-invariance for mim. No complaints on the engineering.

**We're going to park this for post-GA and close the PR** (the branch stays — this reopens cleanly later). The reasoning is scope/timing, not quality:

- **GA focus.** We're cutting the **2.0.0 GA** release now and want to freeze the surface + stop touching the release pipeline. Landing new opt-in modes enlarges the GA payload right as we're trying to shrink it.
- **The payoff is sub-percent.** The Phase 4.0 spike measured convert + decompress at **~0.30 % of wall time**, so a faster convert-step parse — however well done — is a fraction of a percent end-to-end. Nothing here overturns that measurement, so the speed case doesn't clear the bar today.
- **mim's MSRV collides with the GA build.** `mim-input` needs **Rust ≥ 1.91**, but the GA pipeline is pinned to **1.89** (the Docker builder is `rust:1.89-bookworm`, and neither the image nor the release tarball enables `mim-input`). Shipping mim in GA would mean bumping the builder to 1.91 + adding the feature + re-running the release dry-run — release-engineering churn we specifically want to avoid mid-cut.

**To reopen post-GA**, the thing that would move the needle is an **end-to-end wall-clock benchmark** (full pipeline, not a parse microbench) showing a real, reproducible win on a realistic input — or a concrete downstream ask. That's the same posture CBQ is already in (parked on the upstream binseq fix). paraseq alone (1.89-valid, no MSRV snag) would be the lighter half to revisit first if a win shows up.

Really appreciate the contribution — this is parked, not rejected. 🙏

BenjaminDEMAILLE:
Here's the end-to-end benchmark you asked for before reopening this — and the short version is it **confirms your ~0.30 %, so I'm leaving paraseq/mim parked**. None of this touches your GA cut.

**Method.** Two measurements, both reproducible (harness + write-up committed to the branch at `plans/07062026_1025-convert-benchmark/`): (1) the convert step *in isolation* (the only stage `--paraseq`/`--mim` change), interleaved, min+median; (2) the full `bismark … --parallel 1` pipeline via `/usr/bin/time`. Real Illumina `.fq.gz` (1 M + 5 M SE reads); GRCh38 **chr21** bowtie2 index — deliberately small, so Bowtie 2 is *fast* and convert's share is the **largest it can be** (a full genome only shrinks it toward your 0.30 %).

**Convert step, isolated:**

| reads | serial | paraseq | mim |
|------:|:------:|:-------:|:---:|
| 1 M | 0.581 s | 1.15× | **3.57×** |
| 5 M | 3.595 s | 1.10× | **5.01×** |

So mim's parallel gzip decode is genuinely 3.5–5× *on the convert step* — but that's ~0.42 s per 1 M reads in absolute terms.

**End-to-end (1 M vs chr21, interleaved ×3):** serial 14.26 / 10.88 / 10.76 s, paraseq 14.20 / 10.32 / 12.36, mim 12.73 / 9.85 / 13.46. Run-to-run variance is **~30 % (3–4 s)** — an order of magnitude larger than the ~0.4 s convert saving. Convert is ~5 % of wall even on this convert-favourable tiny index, sub-percent on a real genome ⇒ mim projects to **~0.14 %** end-to-end, paraseq **~0.02 %**. No reproducible end-to-end win, exactly as you called it (and a 0.3 % effect isn't resolvable in total wall on a thermally-throttling laptop anyway — the Linux x86_64 host is where a definitive number would come from).

**Verdict:** stays parked; this is the data, not a reopen trigger. The branch is rebased on current `iron-chancellor` and reopens cleanly if a faster-aligner / decompression-bound path ever makes the convert fraction matter.

Separately, re: your "how long to close this out" — the only remaining piece of #1025 with a real case is **CBQ (`.cbq`) input**, which is now unblocked (binseq 0.9.3 fixed the N→A decode, ArcInstitute/binseq#94). I've opened that as its own standalone PR #1052 against `iron-chancellor` so it queues *behind* your release. So: nothing new in GA, CBQ waiting behind it, paraseq/mim parked with data. Freeze away 🙂

PR #1052, opened 2026-07-06
feat(aligner): CBQ (.cbq) read input, un-parked on binseq 0.9.3 (#1025 Phase 3a)

BenjaminDEMAILLE:
## What

Un-parks **CBQ (`.cbq`) read input** for the aligner (#1025 Phase 3a), completing the BINSEQ front-end so both the verbose `.vbq` (Phase 2, shipped) and the columnar `.cbq` variant are accepted. This is the piece that was dropped from #1034 pending an upstream fix.

> **No GA impact — please treat as post-GA.** This targets `rust/iron-chancellor` and queues *behind* the 2.0.0 GA cut. It is additive, opt-in, feature-gated (`binseq-input`, default-OFF / release-ON), and touches nothing on the byte-frozen convert → align → merge core. Land it whenever the release window is clear.

## Why it's unblocked now

CBQ was parked because **binseq 0.9.2 silently mis-decoded `N` → `A`** in release (panic in debug) on the streaming CBQ reader — a no-call turned into fabricated signal, the worst failure class for a methylation caller (ArcInstitute/binseq#94). The upstream maintainer **fixed it in binseq 0.9.3**; this PR bumps the pin `=0.9.2` → `=0.9.3`.

The root cause of the parking was a **test gap**: the CBQ suite inherited the VBQ fixtures, which are ACGT-only, so nothing ever exercised CBQ's headline feature (native `N`). That gap is now closed (see below).

## How

- **Decode path:** the serial streaming `cbq::Reader` (`read_block` → `decompress_columns` → `iter_records`), the deterministic, file-order, **mate-preserving** analog of the VBQ `vbq::MmapReader` path — deliberately **not** `process_parallel` (which reorders blocks and would break R1/R2 pairing). Same "identical to `bqtools decode`" contract, same D2 quality/header fail-loud rejects, same always-compiled extension detection so a feature-off build still rejects `.cbq` never-silently.
- **`N` regression coverage (the fix for the parking):** new `cbq_preserves_n_bases_se` / `_pe` tests assert every `N` survives verbatim at every position (leading / interior / trailing / all-N, both mates). Verified they **FAIL on 0.9.2** (the `sucds` elias_fano panic) and **PASS on 0.9.3** — the regression now has teeth.

## Test plan

- `cargo test -p bismark-aligner` (default / feature-off): **483 passed**; `.cbq` still rejected never-silently.
- `cargo test -p bismark-aligner --features binseq-input`: **484 + 18** (incl. the 8 CBQ analogs + 2 N round-trip tests) passed.
- `cargo clippy -p bismark-aligner --all-targets --release --features binseq-input -- -D warnings`: clean. `cargo fmt --all --check`: clean.
- **End-to-end smoke** (bowtie2, tiny genome, 6 reads incl. 2 bearing `N`): `bismark <reads.cbq>` produces **byte-identical alignment records** to `bismark <reads.fastq>`, and both `N` reads align uniquely with `N` preserved in the aligned SEQ (never silently `A`).

Docs: alignment usage + `rust/README` updated to cover `.cbq` (only `.bq` remains rejected).

Closes the last remaining input-format sub-task of #1025 (paraseq/mim benchmark tracked separately).


FelixKrueger:
Thanks @BenjaminDEMAILLE — this is a careful, correctness-focused PR. We ran a dual code-review and the verdict is **APPROVE on correctness**: the N-regression fix is real and load-bearing, the decode is faithful, and the pin bump is safe. A couple of things to sort before merge, plus one GA-gating follow-up.

### What checked out
- **The N fix has real teeth.** The `cbq_preserves_n_bases_se/_pe` tests byte-assert the full FASTQ across leading / interior / trailing / all-N (both PE mates) — verified to **fail on 0.9.2** (the `sucds` Elias-Fano panic — the debug face of ArcInstitute/binseq#94) and **pass on 0.9.3**. The parking really was a test gap (ACGT-only fixtures); this closes it.
- **Faithful decode.** Serial streaming `cbq::Reader` (file-order, mate-preserving) — not `process_parallel`; the `BlockRange`/`cumulative` arithmetic is inert for the emitted FASTQ (D2 requires headers, so the global index never surfaces). Edge cases covered (empty CBQ via the header-only peek, multi-block, SE/PE, D2 quality/header reject, `.bq` reject, feature-off never-silent).
- **Pin bump is safe workspace-wide.** Diffing binseq 0.9.2→0.9.3, the only functional change is `cbq/core/block.rs` (`resize`→`reserve` + `clear` resets `len_nef`); the **VBQ path is byte-unchanged**, so the merged VBQ backend is unaffected. `=0.9.3` exact-pinned. 👍

### Before merge
1. **Re-target the base branch → `master`.** This targets `rust/iron-chancellor`, which is now fully behind `master` (the GA "master takeover" made `master` the mainline). Please rebase/re-target onto `master` — heads-up that `#1050` (uniform `--version` + per-tool `--help` footer) touched the aligner, so watch for conflicts in `bismark-aligner`.
2. **Confirm `cargo fmt`.** One reviewer's `cargo fmt --all -- --check` (rustfmt 1.9.0) flagged two spots in `tests/binseq_transcode.rs` (~L609 and ~L616). You reported fmt clean, so this is likely a rustfmt-version difference — please run `cargo fmt` against the CI-pinned toolchain and confirm the `fmt` job is green.

### Two doc nits (Low)
- `rust/bismark-aligner/src/lib.rs:292-295` — the `BinseqDecoder` rustdoc still says `.cbq` is *rejected*; this PR now decodes it. Worth updating here since it's the CBQ-enabling change.
- `tests/binseq_transcode.rs:600` — comment says "12 Ns"; the fixture has 11 (the assertion `1+1+1+8` and message are correct — just the prose).

### GA follow-up (not blocking this merge)
CBQ currently has only hermetic round-trip tests — it lacks the real-`bqtools decode` / oxy-scale parity gate that VBQ got. Before we rely on CBQ byte-identity at GA, we'd like: (a) an oxy `bqtools decode` parity cell on a real `.cbq`, and (b) a check that **`bqtools` itself emits records in file order for `.cbq`** — if its own `.cbq` path is the parallel/reordering one, the "identical to `bqtools decode`" contract needs a second look. Happy to run this our side once the branch is re-targeted.

(Optional: the PE desync guard in `cbq_transcode_pe` is structurally unreachable for CBQ — fine to keep for VBQ parity, maybe just note that in the comment.)

BenjaminDEMAILLE:
Thanks for the careful dual review — all points addressed, and I re-based onto `master` so this is ready when you are. Point by point:

**Before merge**
- **Re-targeted → `master`.** Rebased onto current master (the GA takeover). Hit exactly the conflict you flagged: master still carried the old `reject_cbq()` (with the `bismark_rs`→`bismark` rename in its text) while this PR deletes that function — resolved by taking the deletion. #1050 otherwise auto-merged clean.
- **fmt job green.** Root cause was a rustfmt-version skew: CI's fmt job runs toolchain **1.89 (rustfmt 1.8.0)**, my local default is 1.9.0, so I saw "clean" locally and the fmt change never made the first push. Re-ran `rustup run 1.89 cargo fmt --all`; now clean under both 1.8.0 and 1.9.0. `cargo fmt --check` passes ✅ (all 7 checks green).

**Doc nits**
- `lib.rs` `BinseqDecoder` rustdoc now says `.cbq` is **decoded** (Phase 3a), only `.bq` rejected.
- `tests/binseq_transcode.rs`: **11** Ns, not 12 (the `1+1+1+8` assertion + message were already correct).
- While rebasing I also caught a GA-rename miss: `bismark_rs`→`bismark` in the CBQ feature-off message (now matches the VBQ stub).

**Optional**
- Added a comment that the PE `!is_paired()` guard is **structurally unreachable for CBQ** (paired-ness is a file-header property, so every record in a paired `.cbq` carries a mate) — kept for VBQ parity, never-silent regardless.

**GA follow-up (your side, not blocking):** agreed on the oxy `bqtools decode` parity cell on a real `.cbq`. One thing worth confirming there: we decode via the **serial streaming `cbq::Reader`** (deterministic file order), so if `bqtools`' own `.cbq` decode path is the parallel/reordering one, the "identical to bqtools decode" contract is about *content* (same records) rather than *line order* — happy to adjust the wording if your gate shows an order difference. Ready for you to run whenever.

PR #1055, opened 2026-07-06
chore(release): 2.0.1 patch to unblock `cargo install bismark`

FelixKrueger:
## Why

`cargo install bismark` (the whole-suite meta-crate shipped in #1053) is **not installable from crates.io**. Its `src/bin/*.rs` wrappers call `bismark_<tool>::run_main()`, which was added post-GA in #1053 and is **absent from the published GA tool crates** (frozen at `1.0.0` / dedup `1.2.1` *without* `run_main`). A real `cargo publish -p bismark` strips `path=` and resolves those deps from crates.io → `error[E0425]: cannot find function run_main`. (A local `cargo publish --workspace --dry-run` masks this — it resolves the siblings locally, where `run_main` exists.)

crates.io versions are immutable, so the only fix is to republish the tool crates at **new** versions that carry `run_main`.

## What

A **2.0.1 patch** that republishes the 12 tool crates (with `run_main`) at bumped versions, republishes `bismark-meta`, then publishes the `bismark` meta-crate against them.

- `rust/VERSION` + **`rust/bismark-meta/VERSION`** → `2.0.1`. Bumping the vendored copy keeps the `vendored_version_matches_repo_version` drift-guard green **and** makes a registry `cargo install` report `v2.0.1` — the suite version a bare install prints comes from the vendored `VERSION` baked into the published `bismark-meta` tarball (frozen at `2.0.0` in `1.0.0`), not from `rust/VERSION`.
- `bismark-meta` `1.0.0` → **`1.0.1`** (republish). All consumers caret-pin it (`^1.0.0`) → **no pin edits**; it's already first in the publish DAG.
- `bismark-dedup` `1.2.1` → `1.2.2`; the 11 other tool crates `1.0.0` → `1.0.1`.
- **`bismark-io` unchanged** (`1.0.0`) — byte-identical since GA, so no exact-pin cascade.
- `bismark` meta-crate `2.0.0` → `2.0.1` + its 12 exact-pins repointed (`=1.0.1` / dedup `=1.2.2`).
- `release.yml`: fold `bismark` into the publish DAG (`L3`, published last, since it depends on every tool) + widen the Invariant-0 `members` grep to `bismark[a-z0-9-]*` so it captures the bare `bismark` member + the success message `14`→`15`.
- Delete the throwaway `publish-bismark-metacrate.yml` (superseded by folding `bismark` into `release.yml`'s DAG).

**Binaries are byte-identical to GA 2.0.0** — `run_main` is byte-inert (`main.rs` is a thin wrapper over it).

## Validation (local)

- `cargo build -p bismark` → `bismark v2.0.1` compiles against local siblings.
- `cargo build` (default-members) → tool crates build at `1.0.1`.
- `cargo test -p bismark-meta` → **`vendored_version_matches_repo_version ... ok`** + `version_line_shape ... ok`.
- `cargo fmt --all -- --check` → clean.
- `cargo publish --workspace --locked --dry-run --allow-dirty` → all 15 crates packaged + verify-built; cargo-computed order is `io → meta → tools → extractor → bismark` (last); no choke on the already-published `bismark-io 1.0.0`.

The full `cargo test --workspace` runs here in CI (needs bowtie2/minimap2/samtools). The real registry verify-build (the E0425 blind spot) is only exercisable at the real cut → covered by a **post-cut `cargo install bismark → grep v2.0.1` assertion** + a fix-forward runbook (bump `bismark`→2.0.2, re-run).

## Reviewed

Dual independent plan-review before implementation; both reviewers converged on the vendored-`VERSION` / `bismark-meta`-republish requirement (folded in above).

## Follow-up (gated — NOT in this PR)

The actual `release.yml` cut (`dry_run=false` → crates.io publish + `bismark-rust-v2.0.1` tag + `:2.0.1`/`:latest` image + tarballs) is a separate, gated step awaiting maintainer go.

Refs #1053 (install-story W1).


FelixKrueger:
Parking this PR — not merging.

Superseded by the planned **one-crate / one-binary consolidation** (per @ewels architecture feedback): a single `bismark` crate with a clap multicall CLI (`bismark <subcommand>` + `argv[0]` aliases for the classic tool names) fixes `cargo install bismark` *natively and permanently* — a single crate has no cross-crate `run_main` resolution problem, so the E0425 this PR works around cannot occur in that design.

Merging 2.0.1 would republish 13 crates to the immutable crates.io registry days before collapsing 14 crates → 1 — throwaway churn that runs counter to the consolidation goal. `cargo install bismark` stays a 404 in the interim (Docker `:2.0.0`/`:latest`, prebuilt tarballs, and the individual `cargo install bismark-<tool>` crates all work); the unified installer ships with the consolidated release.

Branch kept for reference — the vendored-VERSION / bismark-meta version-mechanism analysis in the description remains valid.

PR #1073, opened 2026-07-14
fix: clean `-V` provenance (no stray "unknown") + no-args shows help

BenjaminDEMAILLE:
Two small CLI/UX fixes for the Rust suite, isolated onto `master` (only the two
bugfix commits, no unrelated work).

## 1. `bismark -V` printed a stray `unknown`

```
bismark (Bismark Rust suite) v3.1.0 (unknown — macos/aarch64 — built 2026-…Z)
```

Any build made outside a `.git` checkout (a bare `cargo install` from a source
tree, a `git archive` tarball) had `build.rs`'s `git rev-parse` fail and fall
straight to the literal `"unknown"`.

- `build.rs` resolves the short hash most-authoritative-first: `$BISMARK_GIT_HASH`
  (CI/Docker override) → `git rev-parse` (checkout) → `.cargo_vcs_info.json`
  (the VCS record `cargo package`/`cargo publish` embeds in the tarball, so a
  registry install reports the real commit) → `unknown`.
- The parenthesised `--version` body is now composed at runtime in
  `meta::version_line`; when the hash is still unavailable the `<hash> — ` prefix
  is dropped, so the line reads `(macos/aarch64 — built …)` rather than surfacing
  a bare `unknown`.

Verified in all three states (vcs-info fallback → real short hash; no git + no
vcs-info → clean line; normal checkout → real hash).

## 2. Running a tool with no arguments printed a terse `error:`

Perl prints a tool's usage/help on a bare invocation; the Rust suite printed a
one-line `error:` and exited 1. A shared `cli::help_if_no_args` now renders the
tool's long help to stderr and returns exit code **2** when only the program
name is present.

Done at each tool's runtime entry point (not clap's `arg_required_else_help`) on
purpose: it keeps `Cli::parse_from([])` returning a defaults `Cli`, which the many
unit tests that parse an empty argv and then assert `validate()`'s "no input"
error depend on.

**Scope:** aligner + `dedup`, `extract`, `bedgraph`, `cov2cyt`, `prepare`,
`bam2nuc`, `nome`, `filter`, `consistency`.

**Deliberately excluded:** `bismark2report` / `bismark2summary` — a bare run there
scans the working directory (auto-discovery) and is a valid mode matching Perl.
Bare `bismark` (no subcommand) is unchanged: the multicall dispatcher still
prints the composed suite help (exit 0). Partial-args paths are untouched, so the
per-tool validation errors still fire when some args are present but input is
missing.

## Byte-identity

No analytical-output path is touched. `-V`/help text and stderr are covered by no
byte-identity gate; only timestamp/hash/help surfaces change.

## Tests

New: `meta::version_body_{includes_hash_when_known,drops_unknown_hash}`; per-tool
`no_args_shows_help` (dedup, extractor) + `align_no_args_shows_help`. Updated: the
`..._no_input_files_errors...` sanity tests pass a flag so they still exercise the
validation error; `align_no_genome_errors` supplies reads-without-genome.

`cargo test -p bismark` green (1426 lib + all integration); `cargo fmt --all --
check` and `cargo clippy --all-targets --release -D warnings` clean.

### DIFF
--- a/rust/bismark/build.rs
+++ b/rust/bismark/build.rs
@@ -17,6 +17,11 @@ use std::env;
 use std::process::Command;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+// The pure `.cargo_vcs_info.json` parser lives in `src/meta/vcs_info.rs` so it is
+// covered by the `cargo test` harness (build scripts are not). We `include!` it
+// here to share the exact logic — `parse_cargo_vcs_short_hash` is defined by this.
+include!("src/meta/vcs_info.rs");
+
 /// Git short-hash resolution, most-authoritative first:
 ///  1. `$BISMARK_GIT_HASH` — an explicit override for CI/Docker builds whose
 ///     context has no `.git` (e.g. a `git archive` tarball).
@@ -54,27 +59,13 @@ fn git_short_hash(manifest_dir: &str) -> String {
     "unknown".to_string()
 }
 
-/// Extract the short (7-char) commit hash from a `.cargo_vcs_info.json` embedded
-/// in a packaged crate. The file (at the crate root) looks like
-/// `{"git":{"sha1":"<40-hex>"},"path_in_vcs":"rust/bismark"}`; we do a minimal
-/// hand-parse (no serde in the build graph) tolerant of whitespace.
+/// Read a `.cargo_vcs_info.json` embedded in a packaged crate (at the crate root)
+/// and extract the short commit hash via the shared
+/// [`parse_cargo_vcs_short_hash`] parser (unit-tested in `src/meta/vcs_info.rs`).
 fn cargo_vcs_short_hash(manifest_dir: &str) -> Option<String> {
     let path = std::path::Path::new(manifest_dir).join(".cargo_vcs_info.json");
     let content = std::fs::read_to_string(path).ok()?;
-    let after_key = content.split("\"sha1\"").nth(1)?;
-    let after_colon = after_key.split(':').nth(1)?;
-    let sha = after_colon
-        .trim()
-        .trim_start_matches('"')
-        .split('"')
-        .next()?
-        .trim();
-    let short: String = sha.chars().take(7).collect();
-    if short.len() == 7 && short.chars().all(|c| c.is_ascii_hexdigit()) {
-        Some(short)
-    } else {
-        None
-    }
+    parse_cargo_vcs_short_hash(&content)
 }
 
 fn build_epoch() -> u64 {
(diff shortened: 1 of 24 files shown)


FelixKrueger:
Thanks for another really clean PR, @BenjaminDEMAILLE 🙏 — I ran it through a two-reviewer pass and both **approve**: no correctness bugs, and byte-identity is intact (the `-V` line is character-identical to the old `VERSION_BODY` when a hash is present, and the `perl-oracle` gate is green). A few non-blocking items we'd love before merge:

**Worth doing:**
1. **Test the other 7 wired tools.** Only `aligner`, `dedup`, and `extractor` have a `no_args_shows_help` test; `bam2nuc`, `bedgraph`, `coverage2cytosine`, `genome prepare`, `nome`, `filter`, and `consistency` are wired but unguarded — a dropped wiring or wrong `about` string in any of them would still pass CI. A cheap `.code(2)` + about-substring assertion each (mirroring the dedup/extractor tests) would lock it in.
2. **Reword one test comment.** `align_no_args_shows_help` says clap's `arg_required_else_help` prints the help, but the implementation deliberately does *not* use that (your `cli.rs` docstring explains why — the empty-argv unit tests depend on the manual approach). The comment could mislead a future maintainer into "simplifying" it into a real regression — please point it at `cli::help_if_no_args`.

**Optional polish:**
3. The `.cargo_vcs_info.json` parser in `build.rs` is the fiddliest new code and can't be unit-tested where it lives (build scripts aren't in the `cargo test` harness). If you lift the pure parsing into a `src/`-side helper, it becomes testable. (We traced it by hand and it degrades safely to `unknown` on every malformed input — so this is durability, not a bug.)
4. Minor: the vcs-info hash is pinned to 7 chars while `git rev-parse --short` can auto-lengthen — purely cosmetic since `-V` isn't byte-gated.

Thanks again — this is a nice UX improvement.

BenjaminDEMAILLE:
Thanks for the thorough two-reviewer pass, @FelixKrueger 🙏 — all four items are addressed and pushed (`440dd6a`, `84cdb2d`):

**1. The other 7 wired tools now have a `no_args_shows_help` test.** `bam2nuc`, `bismark2bedGraph`, `coverage2cytosine`, `bismark_genome_preparation`, `NOMe_filtering`, `filter_non_conversion`, and `methylation_consistency` each get a `.code(2)` + `about`-substring assertion on stderr, mirroring the dedup/extractor tests. A dropped wiring or wrong `about` string in any of them now fails CI.

**2. Reworded the `align_no_args_shows_help` comment.** It now points at `cli::help_if_no_args` and spells out that this is deliberately *not* clap's `arg_required_else_help` (the empty-argv unit tests depend on `Cli::parse_from([])` still yielding a defaults `Cli`), so a future maintainer won't "simplify" it into a regression.

**3. Lifted the `.cargo_vcs_info.json` parser into `src/meta/vcs_info.rs`.** The pure parse now has one home, shared with `build.rs` via `include!`, and is covered by three unit tests: a real vcs-info record, a pretty-printed/whitespace variant, and six malformed inputs that must all degrade to `None` (→ `unknown`). One wrinkle worth flagging: the file uses plain `//` comments, not `//!` — an inner doc comment would be a compile error since `build.rs` `include!`s it mid-module.

**4.** Documented the 7-char pin as git's default `--short` minimum, noting it can be one char shorter than an auto-lengthened checkout build's hash (cosmetic, since `-V` isn't byte-gated).

No analytical-output path touched. `cargo fmt --check` + `cargo clippy --all-targets --release -D warnings` clean; `meta::` lib tests (incl. `vcs_info`) and all eight affected integration binaries green.

Issue #1086, opened 2026-07-31
v2 backends: is a bwa-mem4 (pure-Rust bwa-mem2) backend wanted alongside --rammap?


BenjaminDEMAILLE:
> **Disclosure:** I am the author of bwa-mem4. Filing this as a question, not a pitch — if the answer is "the v2 backend surface is wide enough", that is a fine answer and I will not open a PR.

## Proposal

Add **bwa-mem4** (https://github.com/IPNP-BIPN/bwa-mem4) as a further experimental v2 alignment backend of the Rust `bismark` aligner, alongside the existing `--rammap`.

bwa-mem4 is a native Rust reimplementation of bwa-mem2 (indexer included) whose acceptance criterion is **byte-identical output** vs the reference C binary: same five index files, same SAM bytes. Current status:

| | |
|---|---|
| Index | byte-identical, all five files |
| Single-end | byte-identical on 353,517,767 records |
| Paired-end | byte-identical on 707,312,349 records |
| ALT contigs | byte-identical on GRCh38 analysis set, 261 ALT contigs |
| Speed | SE 2.62x, PE 1.85x vs bwa-mem2 at `-t8` (Apple M4 Max, arm64/NEON; a full x86_64 WGS head-to-head is not yet recorded) |
| Output | SAM, BGZF SAM, BAM, CRAM |

Measured on a real 32.9x human WGS (GIAB HG002, 2x150), not simulated or subsampled.

## Why this is a different lane than rammap

Correction to an earlier framing of mine: rammap also targets identical output to minimap2 — that is its stated goal upstream, so "byte-identity to a C reference" is not what separates the two projects. What separates them is **which aligner, and therefore which reads**:

- `--rammap` gives Bismark a pure-Rust **minimap2**: the long-read / ONT lane (`-x map-ont`, SE-only in Bismark, `--local` / PE / `--combined_index` rejected).
- bwa-mem4 gives it a pure-Rust **bwa-mem2**: short-read WGBS / EM-seq, **single- and paired-end**. Paired-end is the gap — the whole minimap-family backend (minimap2 + rammap) is SE-only in Bismark today.

bwa-mem is also the engine under bwa-meth and BISCUIT, so a bwa-family backend is not exotic for methylation data, even though Bismark has never shipped one.

[...]

PR #1090, opened 2026-08-01
feat(aligner): bwa-mem4 short-read backend + --auto_aligner lane selection (#1086)

BenjaminDEMAILLE:
Closes the question in #1086 with running code, so the decision can be made against a diff rather than a description. **Disclosure: I am the author of bwa-mem4.** If the answer is still "the v2 backend surface is wide enough", this can simply be closed.

## What this adds

**1. `--bwamem4` — a fifth alignment backend (v2, opt-in, concordance-gated).** A subprocess seam onto [bwa-mem4](https://github.com/IPNP-BIPN/bwa-mem4), the pure-Rust bwa-mem2 reimplementation. It covers the **short-read** lane (WGBS, EM-seq) that the minimap family does not: `--minimap2` and `--rammap` are both `map-ont` and single-end.

- argv: `bwa-mem4 mem -t 2 -K 20000000 <index-prefix> <reads>`. `mem` is a subcommand, so it leads the clean-slate option string; the build-then-wipe order of the minimap2 path is kept, so the shared Bowtie 2 base is still validated (`-X` on single-end still dies).
- index: bwa-mem2's **five** side files (`.0123`, `.amb`, `.ann`, `.bwt.2bit.64`, `.pac`), addressed by the **bare prefix** — no `.mmi`-style suffix append. No large-index variant.
- no strand flag is emitted, because bwa has none. Each instance searches both strands and the merge classifies by instance slot, exactly as for the minimap family.
- **SE-only.** Paired-end, `--local` and `--combined_index` are rejected, each naming `--bwamem4` and its actual reason (bwa has no per-strand flag for the PE instance model) rather than borrowing the minimap family's missing-oracle wording.

**2. `bismark_genome_preparation --bwamem4`.** Builds the two converted indices with `bwa-mem4 index <mfa> -p BS_CT|BS_GA`, CT and GA concurrently as before. `--single_fasta`, `--slam` and `--large-index` are rejected (bwa indexes one FASTA; there is no large variant), and more than one converted FASTA in a conversion directory fails loud instead of silently indexing the first.

**3. `--auto_aligner` — pick the lane from the reads.** Samples the first 1000 reads of the input, takes the **median** length and selects `--bwamem4` below `--auto_length_threshold` (default 300 bp, range 50–100000) or `--rammap` at or above it.

```
Note: --auto_aligner sampled 200 read(s) from the first input: median length 100 bp vs threshold 300 bp
      -> --bwamem4 (short-read lane). Override with an explicit aligner flag, or move the boundary with
      --auto_length_threshold.
```

Median rather than mean, so one 40 kb read cannot flip a 150 bp library. It is single-end only (both candidates are), conflicts with every explicit engine flag and with `--illumina_5base`, and an empty, unreadable or truncated input is an **error** — never a silent fallback to Bowtie 2.

## The one behavioural subtlety

bwa **always** emits `XS:i:`, writing `XS:i:0` when it found no suboptimal alignment, where Bowtie 2 simply omits the tag. The merge reads `second_best` as "a competitor exists" (MAPQ ladder + same-thread ambiguity boot), so passing that zero through verbatim would give every uniquely-mapping read a phantom competitor scoring 0. `XS:i:0` is therefore normalized to `None` for this backend only; non-zero `XS` flows through unchanged and every other backend parses verbatim. This is a v2 concordance-model decision, documented next to the code and covered by a test.

## Validation

- `cargo test -p bismark`: **1452 lib + all integration tests green**; `cargo fmt` clean. (`cargo clippy -D warnings` reports pre-existing `byte_char_slices` findings in `bam2nuc/*` under the newer local toolchain; nothing from this diff.)
- 25 new unit tests: index suffix arity, argv shape, the `XS:i:0` rule, detection metadata and version parse, every reject, the `--auto_aligner` guards, and the sniff itself (median, outlier, boundary, truncated FastQ, multi-line FastA).
- **End-to-end smoke run on a 20 kb synthetic genome**, both lanes, driving a `bwa-mem4` CLI shim over the installed `bwa-mem2` (legitimate here precisely because bwa-mem4's contract is byte-identity to bwa-mem2, so the shim exercises the real seam: argv, index arity, SAM tags):
  - `bismark_genome_preparation --bwamem4` → the five side files per conversion directory;
  - `bismark --bwamem4` → 200/200 unique best, strand OT 200, `reads_bismark_bwamem4.bam` with correct `XM`/`XR`/`XG`, report line "Bismark was run with bwa-mem4 … with the specified options: mem -t 2 -K 20000000";
  - `bismark --auto_aligner` on the same 100 bp reads → `--bwamem4`; on 4 kb reads against a minimap2-indexed genome → `--rammap`, 20/20 aligned.
- **Not yet done:** the real-data concordance gate (bwa-mem4 vs bwa-mem2 through the identical Bismark pipeline, on the converted indices, at SAM level, across the three library types). That is the natural Phase-2 gate and I am happy to run it on request — I did not want to spend the compute before the backend itself is wanted.

## Scope kept out deliberately

Paired-end (needs a strand-restriction story bwa does not natively provide), `--combined_index`, and any in-process library backend. All three mirror how `--rammap` was staged.

The Bowtie 2 / HISAT2 / minimap2 / rammap paths and the merge, XM and output core are untouched: every change is a new `Aligner::BwaMem4` match arm, a new flag, or a new guard.


FelixKrueger:
Thanks for this PR. I have thought about this a bit over night, and I don't think that Bismark will need a 5th backend aligner at this time. If I understand this correctly, then bwa-mem4 should not be compared to `minimap2` or `rammap`, but is the equivalent to `Bowtie2`. 

I have been faithful to Bowtie2 for a very long time now, and feel that I have a good understanding of how it works and how it behaves (e.g. with regards to the mapping of N-masked sequences). If folks really want to go down the bwa-route there are options already... Thanks for the suggestion though.

PR #1091, opened 2026-08-01
feat(io): decode gzipped input in parallel via rapidgzip-core

BenjaminDEMAILLE:
## What

Every `.gz` **input** in the suite currently goes through a single-threaded `flate2::read::MultiGzDecoder`, so reading a gzipped FastQ feeds the aligner at one core's worth of inflate no matter how many cores the machine has. This routes all 21 `.gz` input call sites through a new `bismark::io::gzread` module backed by [`rapidgzip-core`](https://crates.io/crates/rapidgzip-core) ([COMBINE-lab/rapidgzip-rust](https://github.com/COMBINE-lab/rapidgzip-rust)), a pure-Rust implementation of the rapidgzip marker/window algorithm.

Full write-up: `plans/08012026_rapidgzip-input/PLAN.md`.

## Byte-identity

Satisfied by construction rather than by argument: the decoded byte stream is fixed by the gzip format and validated by CRC32 + ISIZE, so nothing downstream can observe which decoder produced it. The decode is also thread-budget invariant (unit-tested across 1/2/3/8 threads), so `--parallel` worker-invariance and the Perl v0.25.1 gates are untouched. This is a throughput change only.

## Design

`gzread::open_read` dispatches:

1. no `.gz` suffix, plain `BufReader<File>`;
2. `.gz` + regular file + budget >= 4 + valid framing, `rapidgzip_core::DecoderReader`;
3. otherwise, `MultiGzDecoder` (the previous behaviour).

Step 3 is required, not cosmetic: `ReadAt` needs positional reads over a stable-length source, so a FIFO, a process substitution (`<(zcat x.gz)`) or `/dev/stdin` cannot take the parallel path. It also preserves the pre-existing error text for a non-gzip file named `.gz`.

**Thread budget.** rapidgzip defaults to `available_parallelism()` *per reader*, which is wrong here: `--parallel N` runs N in-process workers, each holding 1 (SE) or 2 (PE) streams open, so the naive default would request `N * streams * cores` decoder threads and starve the aligner it is feeding. The budget is process-global and installed once in `aligner::pipeline`, the single funnel where both the worker count and the layout are known and no stream is open yet. `BISMARK_GUNZIP_THREADS` overrides it suite-wide.

## Measurements

639 MB FastQ decoded, 16-core M4 Max, best of 3. `threads=1` is the sequential `MultiGzDecoder` baseline.

The `budget >= 4` gate is measured, not guessed. With the gate **removed**, rapidgzip is *slower* than the sequential decoder at small budgets, because its speculative decode does redundant work that only amortises once enough workers run:

| threads | `gzip -6` | vs seq | `gzip -1` | vs seq |
|---------|-----------|--------|-----------|--------|
| 1 (seq) | 1048 MB/s | 1.00x  | 675 MB/s  | 1.00x  |
| 2       | 743 MB/s  | **0.71x** | 618 MB/s | **0.92x** |
| 3       | 1111 MB/s | 1.06x  | 911 MB/s  | 1.35x  |
| 4       | 1451 MB/s | 1.38x  | 1213 MB/s | 1.80x  |
| 8       | 1888 MB/s | 1.80x  | 2321 MB/s | 3.44x  |

2 threads is a 29 % regression on ordinary `gzip -6` input, hence `MIN_PARALLEL_THREADS = 4`. With the gate in place (separate run, warm cache, so compare within the run): 1/2/3 collapse onto the sequential baseline as intended, then 1.39x at 4 threads, 2.62x at 8, 3.62x at 12.

## Scope

**In:** `aligner/mod.rs` (6), `aligner/convert.rs` (4), `aligner/genome.rs`, `genome_prep/convert.rs`, `io/genome.rs`, `bam2nuc/genome.rs`, `coverage2cytosine/{cov,genome}.rs`, `nome_filtering/nome.rs`.

**Out, deliberately:**
- Gzip **output** (`GzEncoder` / `gzp::ParCompress`). rapidgzip-core is a decoder only.
- `bedgraph/input.rs`, which uses single-member `GzDecoder` rather than `MultiGzDecoder` to match Perl at that site. Migrating it would change what a concatenated `.gz` yields there.

## Tests

15 new unit tests in `io::gzread`, covering the parity-critical cases: concatenated members are **all** read (the `MultiGzDecoder`-vs-`GzDecoder` distinction the suite relies on), BGZF and single-member round-trip, truncated gzip raising an `io::Error` on read rather than silently returning short output, non-gzip bytes under a `.gz` name failing at read time rather than open time, empty member, missing file, FIFO fallback, and thread-budget invariance.

`cargo fmt --check` clean; `cargo clippy --workspace --all-targets --release` reports nothing in any file this PR touches. Full suite green in debug and release.

Two pre-existing issues on `master`, unrelated to this PR and left alone: `bam2nuc/{report,count,freqs}.rs` produce 23 clippy warnings under current clippy, and `aligner::inprocess::tests::reconstruct_cigar_rejects_pre_clipped_core_in_debug` fails under `--release` because it is `#[should_panic]` on a `debug_assert` that does not fire there.

## Dependency and the risk I want to flag

`rapidgzip-core = "=0.1.0"`, exact-pinned per this crate's convention. Its only dependencies are `crossbeam-deque` and `libz-rs-sys`, i.e. the same zlib-rs inflate backend `flate2` already uses here, so no new system dependency and no C toolchain. MSRV 1.87 sits below the workspace's 1.89. Licensed BSD-3-Clause AND MIT, now recorded in `rust/THIRD-PARTY-NOTICES.md` (BSD-3-Clause was not previously represented).

**Please weigh this before merging:** `rapidgzip-core` 0.1.0 was published on 2026-08-01 with a double-digit download count. This PR enables it by default on every input path, which puts a very new dependency in the critical path of the whole suite with no compile-time opt-out. The mitigations actually in the code are that the sequential fallback survives and is reachable at runtime via `BISMARK_GUNZIP_THREADS=1`, that any construction failure falls back instead of aborting, and that the decode is CRC32 + ISIZE checked. If you would rather have a compile-time switch, say so and I will add the `[features]` gate that `rammap-inprocess` / `binseq-input` already use here; default-on was the shape requested, not a considered preference over a feature flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Bismark Changelog
 
 
+## Unreleased
+
+### Suite-wide
+
+- **Gzipped input is now decompressed in parallel.** Every `.gz` **input** in the suite (FastQ/FastA reads, genome FastA, `.cov.gz`) is decoded through [`rapidgzip-core`](https://github.com/COMBINE-lab/rapidgzip-rust), a pure-Rust implementation of the rapidgzip marker/window algorithm, instead of a single-threaded `MultiGzDecoder`. Reading a gzipped FastQ is no longer a serial bottleneck in front of the aligner. The decoded bytes are unchanged, so this is **byte-identical**: it is a throughput change only. Non-seekable input (a FIFO, a process substitution such as `<(zcat reads.gz)`, `/dev/stdin`) transparently falls back to the previous sequential decoder, as does any input the parallel decoder declines. Gzip **output** (`--gzip`, `.cov.gz`, `.bedGraph.gz`) is untouched.
+- **`BISMARK_GUNZIP_THREADS`** caps the per-stream gzip decoder budget suite-wide. Unset, the budget is `min(8, cores)` divided by the number of read streams `--parallel` can hold open at once, so the decoders never oversubscribe the machine against the aligner they are feeding. A budget below 4 selects the sequential decoder, which is measurably faster at that size, so `BISMARK_GUNZIP_THREADS=1` restores a fully sequential decode without a rebuild.
+
 ## Bismark 3.1.0 (released 2026-07-13)
 
 ### bismark (aligner)
--- a/plans/08012026_rapidgzip-input/PLAN.md
+++ b/plans/08012026_rapidgzip-input/PLAN.md
@@ -1,0 +1,164 @@
+# Parallel gzip input decode via `rapidgzip-core`
+
+Branch: `feat/rapidgzip-input` (off `origin/master`, `ec5cb58`).
+
+Upstream: [COMBINE-lab/rapidgzip-rust](https://github.com/COMBINE-lab/rapidgzip-rust),
+published as [`rapidgzip-core` 0.1.0](https://crates.io/crates/rapidgzip-core).
+
+## Goal and constraint
+
+Reading a gzipped FastQ is currently serial: every `.gz` input in the suite goes
+through a single-threaded `flate2::read::MultiGzDecoder`, so the aligner is fed
+by one core's worth of inflate no matter how many cores the machine has. Replace
+that with `rapidgzip-core`, a pure-Rust implementation of the rapidgzip
+marker/window algorithm, on every `.gz` **input** path.
+
+The constraint is the suite's standing invariant: **byte-identical to Perl
+Bismark v0.25.1**. This change satisfies it trivially rather than by argument,
+because the decoded byte stream is fixed by the gzip format and validated by
+CRC32 + ISIZE. Nothing downstream can observe which decoder produced it. The
+change is throughput-only.
+
+## Scope
+
+**In:** all 21 `.gz` input call sites, via one shared module.
+
+| Area | Sites |
+|------|-------|
+| `aligner/mod.rs` | 6 |
+| `aligner/convert.rs` | 4 |
+| `aligner/genome.rs` | 1 |
+| `genome_prep/convert.rs` | 1 |
+| `io/genome.rs` | 1 |
+| `bam2nuc/genome.rs` | 1 |
+| `coverage2cytosine/{cov,genome}.rs` | 2 |
+| `nome_filtering/nome.rs` | 1 |
+
+**Out, deliberately:**
+
+- **Gzip output.** `GzEncoder` / `gzp::ParCompress` are untouched. `rapidgzip-core`
+  is a decoder only.
+- **`bedgraph/input.rs`.** It uses `GzDecoder` (single-member), not
+  `MultiGzDecoder`, matching Perl's behaviour at that specific site. Routing it
+  through the multi-member decoder would change what a concatenated `.gz` yields
+  there. Left alone on purpose.
+
+## Design
+
+One module, `bismark::io::gzread`, is the single entry point:
+
+```rust
+pub fn open_read(path: &Path) -> io::Result<Box<dyn BufRead + Send>>;
+pub fn open_read_with_threads(path: &Path, threads: usize) -> io::Result<Box<dyn BufRead + Send>>;
+```
+
+Dispatch ladder:
+
+1. no `.gz` suffix, plain `BufReader<File>`;
+2. `.gz` + regular file + budget >= 4 + framing validates, `rapidgzip_core::DecoderReader`;
+3. otherwise, `MultiGzDecoder` (the previous behaviour).
+
+Step 3 is not a nicety. `rapidgzip_core::ReadAt` needs positional reads over a
+stable-length source, so a FIFO, a process substitution (`<(zcat x.gz)`) or
+`/dev/stdin` cannot use the parallel path. Falling back also preserves the
+pre-existing error text for a file named `.gz` that is not gzip, since that error
+is then raised by `MultiGzDecoder` exactly as before.
+
+### Thread budget
+
+`rapidgzip-core` defaults to `available_parallelism()` **per reader**. That is
+wrong here: `--parallel N` runs N in-process workers (`aligner::parallel`), each
+holding open 1 (SE) or 2 (PE) read streams, so the naive default would request
+`N * streams * cores` decoder threads and starve the aligner it is feeding.
+
+The budget is therefore process-global and installed exactly once, in
+`aligner::pipeline` (the single funnel where both the worker count and the layout
+are known, before any stream opens). Resolution order:
+
+1. `BISMARK_GUNZIP_THREADS`;
+2. the installed budget;
+3. `min(8, available_parallelism())`.
+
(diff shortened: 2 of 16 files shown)


FelixKrueger:
Thanks @BenjaminDEMAILLE — this is careful work, and I appreciate you flagging the dependency risk yourself in the PR body rather than leaving me to find it.

I'm going to **park this rather than merge it**, for the same reason as #1034: I measured the ceiling end-to-end and it comes to ~25–45 s of serial inflate on a full 55.7 M-read sample — about 0.1 % of a 10.6-hour run — so there's no reproducible win available at the moment. Separately, `rapidgzip-core` has moved to 0.2.1 since you pinned `=0.1.0`, and one of the things it fixes is a deadlock. Data and the reopen conditions are below. One finding in here I *do* want to chase separately: the `--parallel` splitter inflates the whole input twice per mate, which your PR is what made me go and look at.

<details>
<summary>Supporting data and reasoning (AI-assisted analysis — bin it if it's not useful)</summary>

### 1. The end-to-end ceiling

Decoder throughput measured on a real full-scale input (`SRR24827378_..._R1_val_1.fq.gz`, 1.54 GB compressed → 10.9 GB out, Apple Silicon, warm), then multiplied by the pass count read off the code — not an instrumented aligner run:

| | |
|---|---|
| Sequential decode, one full R1 | 5.9 s (system zlib) / ~10.4 s at the 1048 MB/s `flate2` baseline in your table |
| Full inflates per PE run | **4** — `count_effective` then the split pass, × 2 mates |
| Total serial inflate, 55.7 M-read PE sample | ~25–45 s |
| Share of the profiled 638-min full-sample run | **~0.1 %** |
| Recovered at your best case (2.6× @ 8 threads) | ~15–25 s |

Two structural points behind that. The decode is genuinely serial and *not* overlapped with alignment — `convert_se_files` completes before the aligner is spawned, and it's handed a temp-file path rather than a pipe (`aligner/mod.rs:1168` → `:1214`) — so you're right that this is real serial time and not hidden behind Bowtie 2. But the `drive_merge*` re-reads of the original `.gz` *are* fully overlapped with the subprocess's SAM stdout, so several of the 21 call sites can't contribute anything.

Also worth knowing: sequential zlib on Apple Silicon runs ~1.7 GB/s, i.e. *above* the 1048 MB/s baseline in your table, so some of the measured headroom looks like the zlib-rs backend rather than the parallelism.

### 2. The #1034 precedent

This is the same optimisation through a different crate, and it already has an end-to-end number: `mim` gave 3.5–5× on the convert step in isolation and projected **0.14 % end-to-end**, on a chr21 index chosen specifically to make convert's share as large as it could be — with run-to-run variance around 30 %, two orders of magnitude larger than the effect. Your own conclusion there was to leave it parked. #1004 and #1007 were earlier runs at the same target. Nothing has changed the denominator since; Bowtie 2 is still ~74 % of wall.

The bar #1034 set was: no reopen without an end-to-end measurement. This PR reports only the isolated 639 MB microbenchmark, which is why the bar isn't cleared — not because the microbenchmark is wrong.

### 3. `rapidgzip-core` has moved since `=0.1.0`

| version | date | contents |
|---|---|---|
| 0.1.0 | 2026-08-01 | what this PR pins |
| 0.2.0 | 2026-08-04 | fixes a task-queue race (a worker consuming a task before its counter increment); `DecodeReport` gains mandatory fields — breaking for consumers |
| 0.2.1 | 2026-08-06 | fixes a **deadlock** (all shared workers block while publishing results when the coordinator needs queued resolution work), and parallel-encoder sync flushes forcing an otherwise-suitable stream off the marker/window path |

133 total downloads across all three versions. And unreleased `main` adds a `DecoderPool` with process-wide decode budgets — which would replace this PR's hand-rolled process-global budget and `BISMARK_GUNZIP_THREADS`, i.e. the custom machinery here is about to be obsoleted upstream. Merging now would mean pinning a version with a known deadlock into the critical path of every input, then rewriting the budget layer shortly after.

The 0.2.1 parallel-encoder fix is worth a second look if this comes back: FastQ from `pigz`/`xopen` is common in this field, and on 0.1.0 those streams fall back to sequential — correct, but with none of the speedup, so the measured gains may not reach a lot of real inputs. (I checked my own Trim Galore outputs and they're plain single-stream gzip, so they aren't affected either way.)

### 4. Mechanics

Base is `master`; feature work here goes into `dev` (`dev` → `master` is the release signal). The branch points at `e552b8f`, the 3.1.0 merge, so it predates 22 commits on `dev` including the `legacy_perl/` move — test-merging onto current `dev` conflicts in `CHANGELOG.md`, `rust/bismark/src/aligner/mod.rs` and `.github/workflows/docs.yml`. The green CI run is from 2026-08-01 against pre-`legacy_perl` `master`. License and MSRV are both fine (BSD-3-Clause AND MIT into GPL-3.0-only; 1.87 under our 1.89).

### 5. What reopens this

- A decompression-bound path — a Rust aligner, or in-process `--rammap` — that makes input decode a real fraction of wall clock.
- `rapidgzip-core` maturing past the 0.2.x deadlock/race fixes, with the `DecoderPool` released so the custom budget layer can go away, plus the feature gate you offered (`rammap-inprocess` / `binseq-input` precedent) rather than default-on.
- Either way, an end-to-end measurement on Linux x86_64 rather than a throttling laptop, as you noted yourself in #1034.

### 6. The finding I'd rather chase

Remove a pass instead of speeding one up. `split_contiguous` runs before `std::thread::scope` (`aligner/parallel.rs:666` vs `:675`) and inflates the entire input twice per mate — `count_effective` purely to count reads, then again to write the subsets — single-threaded, with every `--parallel` worker idle. That's a serial fraction that *grows* in relative terms as `N` rises, and halving it needs no new dependency and no byte-identity exposure, provided the chunk boundaries can be derived without the counting pass. I haven't confirmed they can — contiguous chunking may structurally need the total for worker-invariance — so it's a lead rather than a finding.

Unrelated to performance, and consistent with your decision to leave the site alone: `bedgraph/input.rs:65` uses single-member `GzDecoder`, so a multi-member `.txt.gz` would be silently truncated there. Worth confirming separately that the extractor can never emit multi-member output.

</details>

FelixKrueger:
Closing this on the park above. The branch lives on your fork, so nothing is lost and it reopens cleanly if any of the reopen conditions land — `rapidgzip-core` settling past the 0.2.x deadlock/race fixes with `DecoderPool` released, or a genuinely decompression-bound path appearing.

I'm going to take the `--parallel` splitter double-inflate myself. It sits in the chunking path that has to stay worker-invariant, and I want to be the one deciding whether the read count can come out of `count_effective` without breaking that — so please don't pick it up on my account.

Thanks for the work, and particularly for warning me about the dependency yourself instead of letting me discover it. That's the second time your own measurements have made the call for me, and it's appreciated.

Issue #1092, opened 2026-08-01
rammap in-process backend hard-codes the map-ont preset, silently ignoring --mm2_short_reads/--mm2_pacbio


FelixKrueger:
Found while implementing #1081.

## The problem

`--rammap` defaults to the **in-process** backend, and that backend hard-codes the preset:

```rust
::rammap::Aligner::from_index(mmi_str, ::rammap::Preset::MapOnt)   // mod.rs:968
```

(also `inprocess.rs:702`, `:754`). It never reads the preset out of the assembled option string.

The **subprocess** path (`--rammap_subprocess`) does honour it: `options::minimap2_options` fires for `Aligner::Rammap` too, so `-x sr` / `-x map-pb` reach the external binary.

So `--rammap --mm2_short_reads` runs:

| backend | preset actually used |
|---|---|
| in-process (the default) | `map-ont` |
| `--rammap_subprocess` | `sr` |

The flag is accepted, no notice is printed, and nothing warns that it had no effect. That breaks the "same `--mm2_*` knobs, apples-to-apples vs `--minimap2`" contract `rust/README.md` states for rammap, and it makes the two rammap backends non-comparable on short-read data.

## Why it matters more after #1081

Before #1081 the presets differed in alignment scores but Bismark's MAPQ was degenerate (42 for every unique read), so the ignored flag was invisible in the MAPQ column. #1081 makes the denominator `AS`-sensitive, so the silently-ignored preset now moves MAPQ too — `sr` and `map-ont` have different mismatch penalties (8 vs 4) and therefore different `AS` distributions for the same reads.

#1081 records this as a deliberate non-goal (`plans/08012026_minimap2-mapq-denominator/PLAN.md` §1), and its spike documents the finding (`SPIKE.md` F5).

## Options

1. **Plumb the preset through** — map the resolved `--mm2_*` selection to `rammap::Preset` and pass it to `from_index`. Needs a concordance re-check, since it changes in-process alignment for anyone using `--mm2_short_reads`/`--mm2_pacbio` today.
2. **Reject the combination, never-silently** — `--rammap --mm2_short_reads` errors and names `--rammap_subprocess` as the way to get `sr`. Cheapest and fully in keeping with the never-silent rule; leaves the capability gap.
3. Warn and continue — worst of both (a warning nobody reads, plus the gap).

Preference is (1) if the concordance re-check is affordable, otherwise (2) as an immediate stop-gap. What must not stay is the current silent no-op.

[...]

Issue #1081, opened 2026-07-29
minimap2/rammap: positive AS through the end-to-end MAPQ denominator compresses MAPQ to the top rung


FelixKrueger:
Surfaced while fixing #1079. Same defect class, but in a **default** (non-`--local`) path, so it is held byte-frozen rather than fixed alongside.

## The problem

`calc_mapq` normalizes by `diff`, the score range a valid alignment can span. For every mode whose best possible alignment score is 0, that is `abs(score_min)` — which assumes `AS ≤ 0`.

minimap2 and rammap report **positive** `AS:i:` values (the repo's own fixtures note "minimap2-style tags: a positive `AS:i:`"). They route through `calc_mapq` with `local = false`, so:

```
bestOver / diff = (AS + 0.2·len) / (0.2·len) = AS/(0.2·len) + 1  >  1
```

Since every ladder threshold is a fraction of `diff`, the ratio exceeding 1 means the **top rung wins for essentially every unique minimap2/rammap alignment**. MAPQ is therefore near-constant and MAPQ filters (`-q`) are close to no-ops — the same symptom #1079 fixed for Bowtie 2 `--local`.

## Why it was not fixed in #1079

minimap2 SE is **byte-identical to Perl v0.25.1 + minimap2 2.31-r1302** (see `rust/README.md`), so changing its MAPQ deliberately breaks that gate and needs its own parity decision — exactly as the Bowtie 2-local change did.

Note `ScoreModel::from_emitted` currently gives minimap2/rammap `match_bonus = 0.0`. **That is a freeze, not a verified value** (recorded as assumption A9 in `plans/07292026_local-mapq-denominator/PLAN.md`).

## What would settle it

1. Determine minimap2's perfect/best-possible score for the presets Bismark emits (`-x map-ont` default, `-x sr` for 5-Base). Unlike Bowtie 2's `--ma × read_length` this is not a single documented constant.
2. Decide the parity policy: Perl-parity (keep the compressed MAPQ) vs. correctness (match the aligner's own normalization). #1079's precedent was correctness, justified by Bismark's own docs already describing the right formula — there is no equivalent documented formula here.
3. If changing: give minimap2/rammap a nonzero `match_bonus` and revisit whether `ScoreModel::end_to_end`'s aligner-independence still holds (it currently assumes no non-local aligner earns a perfect score).
4. rammap is concordance-gated rather than byte-identical, so it may be able to move ahead of minimap2.

[...]

Issue #787, opened 2026-04-16
Illumina 5-Base Support


Danielsm8:
Hi Felix,

I hope you are doing well! I reviewed some of the prior posts regarding 5-Base and wanted to echo what others have mentioned. Based on my understanding the DRAGEN pipeline does not have flags that will ignore methylation calls in M-Bias regions. This means that the data need to be trimmed, which is not good for cfDNA projects where methylation and fragmentomics features can be concurrently analyzed (this will reduce sequencing coverage, which is already lower). Beyond this, it seems that 5-Base data are locked behind DRAGEN, which really reduces accessibility. Although I imagine this would be challenging, it would be great if someone could develop pipelines that can process and analyze these data without using DRAGEN.

**Update:** I forgot to ask. Can I take the reference used for the DRAGEN processing and plug it into Bismark Methylation Extractor to make the methylation calls with our DRAGEN-processed data?

**Updated 2** It does not work. Any ideas on how to take DRAGEN data and perform independent methylation calls?

Thank you,

Mike

FelixKrueger:
Hi Mike,

I have to say that I am not really familiar with the the 5-Base method from Dragen, and would have to read up around it first. Alternatively, would you have any good documentation around it? I'd also be happy to meet up for a quick Zoom call it that would be easier.

I think I am close to a project to have a go at rewriting the entire Bismark suite soon, maybe adding 5-Base support could be implemented at that point?

Danielsm8:
Hi Felix,

I attached some links below related to 5-Base. I am about to go on leave for an extended period, however, when I get back I'll reach out to you and see if you are available. 5-Base has many advantages over EM-Seq v2, thus I think it would be great if it could be incorporated in an updated Bismark suite.

- https://www.illumina.com/products/by-type/sequencing-kits/library-prep-kits/illumina-5-base-dna-prep.html
- https://www.illumina.com/content/dam/illumina/gcs/assembled-assets/marketing-literature/illumina-5-base-dna-prep-data-sheet-m-gl-03689/illumina-5-base-dna-prep-data-sheet-m-gl-03689.pdf

FelixKrueger:
Thanks for the link, and yes, let's be in touch. All the best!

Danielsm8:
[...]

PR #1096, opened 2026-08-07
Release: 5-Base -> bisulfite-convention converter (#1095) + aligner MAPQ fixes

### LINKED ISSUE #1095: 5-Base → wgbs_tools/UXM interop: re-encode XM into bisulfite-convention SEQ
Follow-up to #787 ([comment](https://github.com/FelixKrueger/Bismark/issues/787#issuecomment-5154564480)). 5-Base data does not work with [`wgbs_tools bam2pat`](https://github.com/nloyfer/wgbs_tools) → [`UXM_deconv`](https://github.com/nloyfer/UXM_deconv) for cell-of-origin deconvolution. The fix is a small standalone converter that writes a second BAM whose `SEQ` has been re-encoded into bisulfite convention from the methylation calls Bismark already made.

## Why `bam2pat` fails on 5-Base data

Bismark's 5-Base methylation is *already* in conventional polarity — the inversion happens when the `XM` tag is written (`aligner/methylation.rs:560-614`), so the extractor, coverage, bedGraph and cytosine reports are all correct as they stand.

`bam2pat` is a Python wrapper around a C++ tool, `patter`, and **`patter` never reads `XM`**. It derives each call from the read character in `SEQ` at CpG positions (`patter.cpp:137-181`), comparing against two constants per strand (`patter.h:59-60`):

```cpp
ReadOrient OT{'C', 'T', 0, 0};   // ref_chr = methylated, unmeth_seq_chr = unmethylated
ReadOrient OB{'G', 'A', 1, 1};
```

5-Base chemistry is the inverse of bisulfite (5mC→T; unmethylated C stays C) and `SEQ` keeps the original read (`aligner/mod.rs:1352-1353`: *"Base-quality masking applies to the CALL ONLY; the BAM SEQ keeps the original read"*). So `patter`'s calls come out **systematically inverted** — not noisy, not degraded.

## What to build

A per-record pure function. No genome, no CIGAR walk, no re-alignment:

```
(meth_chr, unmeth_chr) = XG == CT ? ('C','T') : ('G','A')
ref_base               = XG == CT ?  'C'      :  'G'

for i in 0..len(SEQ):              # len(XM) == len(SEQ), already enforced
    Z X H U  ->  SEQ[i] = meth_chr
    z x h u  ->  SEQ[i] = unmeth_chr
    .        ->  unchanged

NM += (new != ref_base) - (old != ref_base)
MD  = rederive(reconstruct_ref(SEQ_old, MD_old), SEQ_new)
```

Three properties make this work, all verified in-tree:

1. **`len(XM) == len(SEQ)`** is an enforced invariant (parity check in `io::record::from_noodles_record`, relied on at `io/record.rs:301`), and `XM` is reversed in lockstep with `SEQ` for `-`-strand records (`aligner/output.rs:463-467`). So `XM[i]` ↔ `SEQ[i]` positionally in BAM space.
2. **The reference base is knowable from `XG` alone.** Across all four strand indices (`aligner/methylation.rs:131-135`), `methylation_call` emits an `XM` letter only where the genomic base is `C` (CT branch) or `G` (GA branch), and `single_end_sam_output` revcomps `SEQ` and `ref_seq` together for `-` strand. Hence `XG:Z:CT` ⇒ reference base `C` at every letter position; `XG:Z:GA` ⇒ `G`. This is exactly `patter`'s `ref_chr`, so `NM`/`MD` are recomputable **exactly, without the genome**.
3. **Soft clips and insertions are structurally `.`** — `I`/`S` pad the genomic window with `b'X'` without advancing `pos` (`aligner/methylation.rs:174-181` SE, `:349-352` PE `walk_mate`). `X` matches no read base, so `methylation_call` emits `.` there. The window is therefore never mis-framed, and a positional zip provably cannot touch a clipped or inserted base.

### ⚠️ Implementation traps

- **Do not use `iter_aligned()`.** Property 3 removes any need for reference positions, and `iter_aligned()` returns 5'-*oriented* positions — writing into `SEQ` at `read_pos_5p` silently reverses every OB read's edits. See the existing warning at `extractor/call.rs:183`.
- **`U`/`u` must be in the rewrite set.** `push_ct_context` maps an `X` in the context slots to `U`/`u`, so a cytosine immediately before a soft clip or insertion is classified unknown-context. Skipping `U`/`u` leaves exactly those cytosines in 5-Base polarity while their neighbours flip.
- **Fail loud if `MD` is absent** rather than guessing; Bismark always writes it (`aligner/output.rs:485-487`).

### Consumer requirements (from `bam2pat.py`)

Coordinate-sorted **and** indexed (`is_bam_sorted`), `-q 10` (`MAPQ = 10`), `-F 1796` (`FLAGS_FILTER`), and `-f 3` for paired-end. FLAG vocabulary is `{99,147}` / `{83,163}` (PE) or `{0}` / `{16}` (SE), and `is_bottom` (`patter_utils.cpp:163-168`) agrees with Bismark's `XG` on every one. So `samtools sort` + `index` is the only gap.

## Decisions taken

**Converter, not direct `.pat` emission.** `.pat` is keyed on wgbs_tools' own CpG index — `locus2CpGIndex` (`patter.cpp:79-91`) looks up a locus in a dict built by *their* `init_genome`, per genome build, and throws `std::logic_error("Reference Error. Unknown CpG locus")` on any mismatch. Emitting it would also mean reimplementing their read pairing, `clip_size`, `min_cpg`, filters and pat-collapse. Decisively: the scientific goal is that paired 5-Base and EM-seq arms stay comparable — a converted BAM enters at the *front* of the existing path so both arms pass the same filters, whereas `.pat` injects 5-Base *past* every filter.

**A flag on the aligner, `--five_base_bisulfite_bam <BAM>`**, dispatched early exactly like `--five_base_consensus_from_bam` (`aligner/mod.rs:167` → `:515`). Keeps all `--five_base_*` surface in one place and avoids committing to a new permanent subcommand plus classic-name alias in `cli.rs:54-72`.

**Convert all cytosine contexts, not CpG only.** `patter` gates on a CpG mask so all-contexts costs it nothing; restricting to CpG would mean *adding* a filter; and a half-converted file mixes two polarities silently for every other reader.

## Non-goals and caveats

- **Never the primary BAM.** `SEQ` no longer matches what the sequencer read. Separate, clearly-named output.
- **Lossy at cytosines.** The re-encode replaces observed bases with called ones. Since 5-Base cannot separate a genuine `C>T` from 5mC without the opposite strand, real variants get written as methylation — precisely the sites `--five_base_deconvolution` exists to find. Harmless for UXM (fragment-level CpG patterns); the converted file must never be used for variant calling. Optional interaction worth considering: mask deconvolution-called variant sites to `N`.
- Naming clash to keep in mind: `--five_base_deconvolution` means *variant-vs-methylation*, **not** cell-of-origin.

## Confirmed inputs (no longer blocked)

Answered by the reporter in [#787](https://github.com/FelixKrueger/Bismark/issues/787#issuecomment-5167178704) on 2026-08-03:

1. **Existing BAMs are from DRAGEN, but he is happy to re-run through Bismark.** So the XM-driven design above stands — no genome-driven variant needed.
2. **`bam2pat` defaults only**, plus `--clip` for M-bias. Genome is hg38 via `wgbstools init_genome hg38`.
3. **Paired-end** — consistent with 5-Base being PE-only.
4. Variant/methylation deconvolution is not needed at his 10X coverage (possibly at 100X, "not anytime soon"), so the lossy-at-cytosines caveat is acceptable and masking variant sites to `N` is deferrable.

⚠️ **Chromosome naming.** His wgbstools genome is `init_genome hg38`, so the Bismark genome he re-aligns against must use the same naming (`chr1`-style). `locus2CpGIndex` throws `std::logic_error` on an unknown locus — loud, but only at the end of a long run.

## Alternative worth knowing: patch `patter` instead

`patter` can be made natively 5-Base-correct by swapping two constants. `ref_chr`/`unmeth_seq_chr` are read in **exactly two places** (`patter.cpp:153`, `:159`) and nowhere else, so

```cpp
ReadOrient OT{'T', 'C', 0, 0};   // was {'C','T',0,0}
ReadOrient OB{'A', 'G', 1, 1};   // was {'G','A',1,1}
```

inverts the call polarity completely. It survives `is_cpg()`, which accepts `C|T` (OT) and `G|A` (OB) and whose required context bases (`seq[j+1]=='G'`, `seq[j-1]=='C'`) are guanines on the read's own strand in 5-Base chemistry and therefore never converted.

This works on **any** 5-Base BAM including DRAGEN's, with no re-alignment and no Bismark involvement. Tracked as a possible upstream contribution to `nloyfer/wgbs_tools`; it does not replace this issue, since the converter serves Bismark 5-Base users without requiring a patched third-party binary.

**Ruled out — `MM`/`ML` modification tags.** `patter` already auto-detects standard SAM `MM`/`ML` tags (`patter.cpp:334-338`) and has a modification-aware path, which would have avoided falsifying `SEQ` entirely. But `patter.cpp:341-343` throws `"Unrecognized bam format: paired end and nanopore"`, and 5-Base is paired-end only. Blocked unless upstream lifts that restriction.

**Precision note:** `patter`'s *call* path never reads `XM`, but `is_pass_ds_test` (`patter_utils.cpp:357-362`) does parse `XM:Z:` for the opt-in `--ds_test` filter. Does not affect the diagnosis or the fix.

FelixKrueger:
Release merge of `dev` into `master`: 22 commits covering one new aligner feature, four MAPQ/CLI fixes, and the supporting plans and fixtures. Per this repo's convention a `dev`→`master` merge **is** the release signal, so this also closes the linked issues.

> ⚠️ **Not merge-ready yet — the version has not been bumped.** All four mirrors still read `3.1.0` (`rust/VERSION`, `rust/bismark/VERSION`, `rust/bismark/Cargo.toml`, `rust/README.md`) and `CHANGELOG.md` still has `## Unreleased`. Magnitude is a **minor bump → 3.2.0**: several changes alter behaviour in *default* paths (see MAPQ below). Bump those five places, then merge.

## New: `--five_base_bisulfite_bam` (#1095, from #787)

`wgbs_tools`' `bam2pat` — and so `UXM_deconv` cell-of-origin deconvolution — derives methylation from the read base in `SEQ` at each CpG and never reads `XM`. Illumina 5-Base `SEQ` holds the raw read and 5-Base chemistry is bisulfite's inverse, so those calls came out **systematically inverted**: not noisy, inverted.

The new flag re-encodes an existing 5-Base BAM's `SEQ` into bisulfite convention from the `XM` tag Bismark already wrote. No genome, no re-alignment: the reference is rebuilt per record from `MD` and *proved* against that record's own `NM`/`MD` before anything is emitted.

**Measured end-to-end with stock, unmodified `patter`** on a fully CpG-methylated 5-Base control:

| Route | METH / total | Methylation |
|---|---|---|
| stock `patter` + raw 5-Base BAM | 0 / 152 | **0.0 %** |
| patched `patter` + raw 5-Base BAM | 152 / 152 | 100.0 % |
| **stock `patter` + converted BAM** | **152 / 152** | **100.0 %** |

Ground truth is 100 %; Bismark's own `XM` agrees (`Z=168, z=0`). The reporter has since [confirmed it on real data](https://github.com/FelixKrueger/Bismark/issues/787#issuecomment-5219193146).

Safety properties worth knowing before release:

- **Never the primary BAM.** `SEQ` no longer matches the sequencer, and genuine `C>T` variants are encoded as methylation — fine for fragment-level CpG tools, not for variant calling. Stated in the CLI help, the docs and the CHANGELOG.
- **No tunable behaviour at all.** One new flag; the encoding and the `N`-masking of no-call cytosines are both fully determined by the record.
- **Refused runs leave no output.** `BamWriter`'s `Drop` writes the BGZF EOF marker, so an abandoned file would pass `samtools quickcheck` and read as a finished conversion; every error path now removes the output and the report.
- The reported **flip rate** is exactly `1.000` for 5-Base input and `0.000` for bisulfite; anything between fails the run. Converting an ordinary bisulfite BAM is therefore a no-op by construction, and that identity is a CI gate.

## Aligner MAPQ fixes — these change default-path output

- **#1081** — minimap2/rammap were normalized by `abs(score_min)` despite reporting *positive* scores, so **MAPQ was 42 for every uniquely-aligned read**. Now uses `2 × read_length` and the local ladder. **Retires minimap2 SE's byte-identity vs Perl v0.25.1**, deliberately — the old values were informationless.
- **#1083** — Bowtie 2 `--local` gets Bowtie 2's own `max(1, perfect_score − score_min)` denominator. End-to-end stays byte-identical.
- **#1088** — HISAT2 `--local` uses the **end-to-end** ladder and the linear score-min form it is actually emitted. Its MAPQ floor can now reach 0, so `-q` filters downstream of `--hisat2 --local` will drop reads they previously kept.
- **#1092** — `--rammap` now honours `--mm2_short_reads`/`--mm2_pacbio`; the in-process backend had hard-coded `map-ont`, so the flag was accepted and ignored. **Changes mapping rate, not just scores.**

## Also included

- **#1085** short options accepted in either letter case, as Perl's `Getopt::Long` did · **#1082** score-min trio collapsed into one `ScoreModel` · **#1075** `-p` routed to minimap2/rammap `-t` · **#1073** clean `-V` provenance + no-args help · **#1072** Homebrew install docs + docs build on PRs
- New oracle-validated soft-clip fixture (`rust/bismark/tests/data/five_base_bisulfite/`) — **no fixture in the repo previously contained a single `S` op**, while `nS`-prefixed reads are the dominant real 5-Base CIGAR shape
- `samtools` added to the CI `test` job, and the #1095 gates now **panic rather than skip** when it is absent — three gates, including the load-bearing idempotence gate, had been reporting green in CI while asserting nothing

## Verification

`cargo fmt --check` clean · `cargo clippy --all-targets` clean · **79 test suites, 0 failures**. #1095 specifically: 25 unit tests + 7 integration gates, and its deletion path is validated against an *independent* from-genome `MD`/`NM` oracle (`3S5M2D4M1I4M1D3M` → `5^GA4G1A1^T0G1A0`, `NM 11`) rather than against itself.

#1095 went through the full pipeline in one session: plan rev 0→2, dual plan review plus a targeted section review, implementation, then dual code review plus a Mode-B coverage audit. All reports are committed under `plans/08062026_five-base-bisulfite-bam/`. **No Critical findings**; both code reviewers confirmed the algorithm correct and independently verified byte-identity at scale (12 974 records, and all four strand indices).

**Known gaps, all in validation and none in behaviour**, recorded in `PLAN.md` §10b and the handoff's G9: the unmapped/secondary pass-through has no committed test (no tracked fixture has unmapped records *and* `MD`; behaviour was verified during review), the `XG` ⟺ FLAG equivalence is documented but not asserted, committed tests are single-end covering 2 of 4 strand indices, and a lower-case `MD` base from a non-Bismark BAM would silently miss a mask.

Closes #1095.


FelixKrueger:
Closing as premature — this was opened as a `dev`→`master` merge, which in this repo *is* the release signal, and there is no release pending. Work is still settling and new issues are still arriving; the release should be a deliberate cut at a chosen point, not a by-product of landing #1095.

Nothing is lost: #1095 and everything else stays on `dev`, and reopening this PR restores the body verbatim when a 3.2.0 cut is actually wanted. The five version literals still read `3.1.0`, which is correct for now.

PR #1115, opened 2026-08-29
feat(extractor): opt-in --allow_discordant for aligner-agnostic input

nh13:
## Summary

Adds an opt-in `--allow_discordant` flag to `bismark_methylation_extractor` so it can call methylation on BAMs from **any** bisulfite aligner that emits Bismark-format `XM`/`XR`/`XG` tags — not only the Bismark aligner. General-purpose aligners (e.g. bwa-mem3 `--meth`) emit records the Bismark aligner never does — cross-chromosome and discordant-orientation pairs, orphans whose mate is unmapped, and secondary/supplementary alignments — on which the current extractor hard-errors, forcing a `samtools view -f 0x2 -F 0x900` pre-filter. This flag lets those reads be extracted directly.

**Off by default, and byte-identical to current behaviour when off** (including today's cross-chromosome and unpaired-record errors) — every new code path is gated behind the flag.

## What the flag does

- **Skips + counts** secondary (`0x100`) / supplementary (`0x800`) alignments at the reader layer, before record validation and pairing.
- **Classifies** each adjacent primary pair by strand: a genuine Bismark concordant pair (same chromosome, complementary orientation, **same strand of origin** — `{OT,CTOT}` or `{OB,CTOB}`, order-agnostic so non-directional libraries still match — and consistent geometry) takes today's concordant path verbatim. Everything else — cross-chromosome, same-orientation (FF/RR), opposite-strand-of-origin, or annihilated-geometry pairs — is called **independently**, each mate by its own record strand against its own chromosome.
- **Tolerates orphans** (a mapped primary whose mate was unmapped) via a qname pre-compare + one-slot pushback, calling them single-end style.
- **Deduplicates** same-chromosome overlapping independent pairs by reference position (keeping R1's calls) rather than reusing the FR-only half-plane `drop_overlap`, so shared cytosines are counted exactly once and never mis-dropped.
- **Reports** every class in a dedicated splitting-report section, and warns (not gated by `--quiet`) when >20% of records are orphans — a sign the input is not name-grouped.

## Validation

New `tests/extractor_allow_discordant.rs` covers every class (cross-chr; same-chr disjoint and overlapping; opposite-strand-of-origin overlap; orphans with both R1- and R2-survivors; secondary/supplementary skips; SE mode), plus a **flag-on ≡ flag-off byte-identity lock** on concordant input, a `--parallel 1 ≡ 4` invariant, and a single-threaded ≡ parallel oracle. Parallel and single-threaded pipelines share one `classify_pair`, so both get the same behaviour from one implementation.

## Follow-up: docs (preview)

I've drafted the documentation update as a separate, **stacked preview PR** on my fork so it doesn't clutter this review: nh13/Bismark#1. It documents `--allow_discordant` in the options reference and adds an "Input from a non-Bismark aligner" section to the methylation-extraction guide, including the measured trade-off of aligning with a general-purpose bisulfite aligner ([bwa-mem3](https://github.com/fg-labs/bwa-mem3) `--meth`) and extracting through `bismark extract --allow_discordant`: ~6.5× faster alignment on a human WGBS slice, per-CpG methylation concordance _r_ ≈ 0.97 with matching specificity, with the Bismark aligner keeping a small per-read sensitivity edge and the general-purpose aligner reporting higher yield.

It's framed as a starting point rather than a blanket recommendation. Happy to open it as a proper PR here once the feature lands, or to reword/drop the recommendation if you'd rather keep aligner suggestions out of the docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Bismark Changelog
 
 
+## Bismark (unreleased)
+
+### bismark extract (methylation extractor)
+
+- **New opt-in `--allow_discordant` for aligner-agnostic input.** The methylation extractor can now consume BAMs from general-purpose bisulfite aligners that emit Bismark-format `XM`/`XR`/`XG` tags — not just the Bismark aligner — without an external `samtools view -f 0x2 -F 0x900` pre-filter. With the flag it skips (and counts) secondary (`0x100`) / supplementary (`0x800`) alignments, calls cross-chromosome and same-chromosome discordant pairs independently, and calls orphan reads whose mate is unmapped; every new class is tallied in a dedicated splitting-report section. It is **OFF by default and byte-identical to the previous behaviour when off** (including the historical cross-chromosome / unpaired-record errors). A stderr warning (not gated by `--quiet`) fires when more than 20% of records are orphans, which usually means the input is not name-grouped (query-sorted / name-collated).
+
+
 ## Bismark 3.1.0 (released 2026-07-13)
 
 ### bismark (aligner)
--- a/rust/bismark/src/extractor/cli.rs
+++ b/rust/bismark/src/extractor/cli.rs
@@ -212,6 +212,22 @@ pub struct Cli {
     #[arg(long = "mbias_off")]
     pub mbias_off: bool,
 
+    // ─── Aligner-agnostic discordant handling ───
+    /// Opt-in: call methylation on records a Bismark aligner never emits, so
+    /// output from a general-purpose bisulfite aligner (emitting Bismark-format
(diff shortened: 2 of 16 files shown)


FelixKrueger:
Hi Nils, 

Thanks for this contribution. Conceptually I think would be a fitting addition that would allow `bwa-mem3` output to feed into the methylation extraction (and beyond). I am currently vacationing and trying to keep computer work to a minimum, so apologies of review and merging will take some time. 

In the meantime, could you please rebase the PR to target the `dev` branch instead of master as it is reserved for releases (which will shortly after I am back at work). Cheers

PR #1121, opened 2026-09-11
aligner: stream the converted reads to the aligner instead of writing them (#1120)

### LINKED ISSUE #1120: feat(aligner): FIFO streaming for the converted temp reads (disk-constrained runs)
## Summary

The aligner writes the in-silico converted reads (`_C_to_T` / `_G_to_A`) to `--temp_dir` on every run, and `--multicore` adds the chunk subsets (`.temp.N`) on top. For users on constrained scratch this is the binding constraint: the run cannot start, however much CPU or RAM is available. `--gzip` takes roughly 5x off but never reaches zero.

Bowtie 2 and minimap2 both accept FIFOs for their read inputs and produce byte-identical alignment records, so the converted reads can be streamed instead of written. This proposes streaming as the default path, with the existing file path retained as a fallback. It measures as slightly faster than the current path, not slower, because it takes conversion off the critical path.

Targets the Rust suite (`upstream/dev` @ `170bfce`), per the Perl maintenance freeze in https://github.com/FelixKrueger/Bismark/issues/1010.

Same problem class as https://github.com/FelixKrueger/Bismark/issues/1019 (BGZF spill compression, shipped in PR https://github.com/FelixKrueger/Bismark/pull/1022), one layer up: that issue made a scratch file smaller, this one removes a scratch file.

## Current disk cost

Directional PE, the converted FastQ is roughly the uncompressed size of the input per converted set. On this repo's `test_files` (5,000 read pairs, 436 KB gzipped across R1+R2) the converted pair is 2.3 MB — 5.3x the input as delivered.

| | `--multicore 1` | `--multicore N` |
|---|---|---|
| `.temp.N` subsets | none | ~1x uncompressed input, always plain |
| converted FastQ | ~1x uncompressed input | ~1x uncompressed input |
| per-chunk BAMs | none | ~1x output |

All N chunks are live at once — workers spawn together under `std::thread::scope` (`rust/bismark/src/aligner/parallel.rs:675`) and each deletes its subset and converted files only after its own alignment finishes (`parallel.rs:440-444`), so per-chunk cleanup does not lower the peak.

`--multicore 1 -p N` is already the right advice for unrelated reasons (the benchmarks page has pure `-p` both faster and 5.7x lighter on RAM than pure `--multicore`) and it does remove the subsets. It still leaves one converted pair on disk, and there is currently no way to avoid that.

## Why streaming is tractable

Three properties of the existing code make this contained rather than a redesign.

**Conversion is a pure per-read map.** `convert_seq_c_to_t` / `convert_seq_g_to_a` (`aligner/convert.rs:103`, `:149`) are stateless — no index, no lookback, no second pass.

**The aligner is the only consumer of the converted bytes.** The methylation caller re-reads the *original* FastQ in lockstep (`aligner/mod.rs:3084`; `convert.rs` header: "The *original* (unconverted) read is deliberately NOT retained here — it is re-read in lockstep during the later methylation-call loop"). Nothing needs the converted file once the aligner has consumed it.

**The temp path cannot reach the output.** Bismark parses only Bowtie 2's alignment records and writes its own header, `@PG ID:Bismark ... CL:"bismark <argv>"` (`aligner/output.rs:107`); Bowtie 2's own `@PG` is not propagated. So byte-identity here holds by construction, not by argument: the bytes the aligner receives are identical, and the one SAM field that records the input path is discarded.

The reason it is a file today is that each converted file has exactly **two** concurrent readers, in every library mode — `se_instance_plan` (`mod.rs:1151`) is `[(Norc, Ct, 0), (Nofw, Ga, 0)]` for directional, both instances reading file 0, and `0,0,1,1` for non-directional; `pe_instance_plan` (`mod.rs:4733`) has directional slots s0 and s3 "both read `-1 C→T_R1 -2 G→A_R2`". Two readers rule out a single pipe. They do not rule out one pipe per reader.

## Proposed design

One FIFO per aligner instance per mate. One conversion pass per mate, fanned out to that mate's FIFOs through a bounded queue, one writer thread per FIFO. The argv shape stays exactly `-1 <path> -2 <path>` (`aligner/align.rs:261`), so nothing about how the aligner is invoked changes.

Directional PE is 2 instances x 2 mates = 4 FIFOs fed by 2 conversion passes. Non-directional is 4 instances = 8 FIFOs, still 2 passes (the C→T and G→A sets derive from the same two inputs).

CPU is one conversion pass per mate, the same as today. Memory is bounded by queue depth, not by read count. No re-read of the input is needed, so non-seekable inputs are unaffected.

Two constraints came out of testing and are not optional:

**One writer per mate is mandatory.** A single writer that fills mate 1's FIFO before starting mate 2 deadlocks — the aligner blocks reading mate 2 while the writer blocks on mate 1's 64 KB pipe buffer. Confirmed (timeout, exit 124).

**HISAT2 cannot be served this way** (see below), so it falls back to the file path under `--hisat2` — never silently.

## Validation

Bowtie 2 2.5.5, HISAT2 2.2.2, minimap2 2.31 — the versions the repo Dockerfile pins for byte-identity. Genome and reads are this repo's `test_files` (E. coli K12 `NC_010473`, 5,000 read pairs) against a C→T-converted index, 51.26% overall alignment rate on the `--norc` instance so the alignment path is genuinely exercised. Comparisons exclude the `@PG` line only, since it necessarily records the input path and Bismark does not propagate it.

| Test | Result |
|---|---|
| Bowtie 2 PE, FIFO per mate | byte-identical |
| Bowtie 2 PE, FIFO fed straight from the converter (zero converted bytes on disk) | byte-identical |
| Bowtie 2 SE, `-U` over a FIFO | byte-identical |
| Both directional instances concurrently, independent FIFO pairs | byte-identical (both) |
| Gzipped bytes over a FIFO | byte-identical (gz is sniffed from the pipe) |
| `-p 4 --reorder` over FIFOs, as Bismark invokes it | byte-identical |
| Tee fan-out, one pass to both instances, queue depth 1 / 4 / 64 blocks | byte-identical at every depth |
| Tee fan-out with asymmetric consumers (`-p 1` vs `-p 4`, depth 1) | byte-identical, no starvation |
| minimap2 SE over a FIFO | byte-identical |
| **HISAT2 PE and SE over a FIFO** | **fails, exit 255** |
| Single writer filling mate 1 before mate 2 | deadlock (exit 124) |

## Performance

Conversion currently runs to completion before the aligner is spawned (`process_se_chunk` calls `convert_se_files` then spawns — `mod.rs:1200`), so it sits on the critical path. Streaming overlaps it with alignment. 300,000 read pairs, both directional instances, `-p 4`, three reps:

| | convert | align | total | converted files on disk |
|---|---|---|---|---|
| Serial (current shape) | 472-605 ms | 4148-4210 ms | 4621-4815 ms | 139 MB |
| Streamed | overlapped | — | 4174-4355 ms | 0 |

About 10% faster here. FIFO transport itself showed no measurable overhead in isolation (106-110 ms vs 107-109 ms, `-p 4`, mean of 5).

Treat the 10% as an upper bound rather than a forecast: the stand-in converter is `awk`, and the real Rust converter is faster, which shrinks the serial phase being hidden; and E. coli makes alignment cheap relative to a human index. The defensible claim is **wall-neutral to modestly faster, never slower** — the change removes a serial phase rather than adding work.

## Suggested scope

Streaming becomes the default path for Bowtie 2 and minimap2, with `--no_stream_converted` (name up for grabs) as an escape hatch that forces the current file path — needed to A/B the two paths in tests and to bisect if something unexpected appears.

The file path is retained, not removed, and is selected automatically when streaming is unavailable: under `--hisat2`, and when `mkfifo` is unavailable or `--temp_dir` cannot host a FIFO (network scratch is exactly where users point `--temp_dir`). That fallback must be never-silent — a silent one would hide the disk cost the feature exists to remove.

Default-on rather than opt-in on the understanding that full-scale byte-identity validation gates the merge; the flag exists so that validation can compare both paths directly. `--combined_index_sequential` is the precedent for a mode that landed opt-in and later became the default with the flag "retained as an explicit selector" (`cli.rs:268`) — this starts one step further along, because the file path never goes away.

Out of scope, deliberately:

**The `.temp.N` subsets.** `--multicore 1 -p N` already avoids them and is the recommended configuration anyway.

**The uBAM / BINSEQ transcode FastQ** (`mod.rs:295-310`), which is consumed twice — by the convert step and the methylation-call re-read — so it genuinely has to be a file.

**Making `--gzip` parallel.** It currently uses single-threaded `flate2::GzEncoder` at `Compression::default()` (`convert.rs:284`, `:469`, `:591`), and `gzp` / `noodles-bgzf` are already dependencies, so a faster writer would be cheap. But it does not help the disk-constrained case — a smaller file is still a file — and if streaming lands its only remaining consumer is HISAT2. Noted, not proposed.

## Two things to record while they are in view

Neither is a work item here.

**HISAT2 has no path to zero.** It rejects a FIFO for both `-1/-2` and `-U`:

```
Warning: Unsupported file format
(ERR): Read file 'h1' doesn't exist
Exiting now ...
```

It appears to stat or seek the input during format sniffing. Since `--hisat2` already reinterprets `--multicore N` as single-instance `-p N` threading (`aligner/cli.rs:57`) and so writes no subsets, HISAT2 users are left with one converted set on disk and `--gzip` as their only lever.

**`--gzip` does not reach the `.temp.N` subsets.** They are written plain by design (`parallel.rs:151`: "NO prefix, NO `.gz` — written plain so the converter's suffix-based gz detection reads them plain"), where Perl compressed them when `--gzip` was set (`bismark:139`). So `--gzip --multicore N` currently costs an extra full uncompressed copy of the input relative to Perl. Flagging it only — that is the chunking path PR https://github.com/FelixKrueger/Bismark/pull/1091's closing comment reserves, and it needs to stay worker-invariant.

<details>
<summary>Reproduction</summary>

Container, using the Dockerfile's pinned versions:

```dockerfile
FROM mambaorg/micromamba:1.5.8-bookworm-slim
USER root
RUN micromamba install -y -n base -c bioconda -c conda-forge \
      bowtie2=2.5.5 hisat2=2.2.2 minimap2=2.31 python=3.12 \
    && micromamba clean --all --yes
ENV PATH=/opt/conda/bin:$PATH
WORKDIR /work
```

Inputs are `test_files/NC_010473.fa.gz`, `test_files/test_R1.fastq.gz` and `test_files/test_R2.fastq.gz` from this repo. The genome is C→T converted; reads are C→T (R1) and G→A (R2) converted; Bowtie 2 gets a Bismark-shaped option set (`-q --phred33 -L 20 --score-min L,0,-0.2 --ignore-quals --no-mixed --no-discordant --dovetail --minins 0 --maxins 500`) plus `--norc` / `--nofw` per instance. The core check:

```bash
mkfifo c1 c2
convert R1.fq.gz c1 ct &   # streams converted records into the FIFO
convert R2.fq.gz c2 ga &
bowtie2 $OPTS --norc -x idx -1 c1 -2 c2 -S fifo.sam
cmp <(grep -v '^@PG' file.sam) <(grep -v '^@PG' fifo.sam)
```

Full scripts (including the tee fan-out with bounded queues, and the timing harness) available on request.
</details>

ewels:
Closes #1120.

## The problem

Bismark converts your reads in silico before aligning them, and every run wrote that converted set to `--temp_dir` first — roughly 1x the uncompressed input per converted set, and up to four sets for a non-directional paired-end run. For users on constrained scratch that was the binding constraint: not slow, but unable to start at all. `--gzip` took roughly 5x off it and never took it to zero.

Those bytes have exactly one consumer, the aligner. The methylation call re-reads the *original* FastQ, so nothing else needs them. They are now handed to the aligner through named pipes and never touch the disk.

## Measured

200,000 read pairs (93 MB uncompressed across R1+R2), directional paired-end against the E. coli index, `-p 4`, release build, 7 reps with the two arms **interleaved** so any thermal drift is shared rather than handed to one arm.

| | wall, median of 7 | peak converted bytes in `--temp_dir` |
|---|---|---|
| Files (`--no_stream_converted`) | 6.35 s | **96 MB** |
| Streamed (default) | 6.33 s | **4 KB** |

The 4 KB is the FIFO inodes and does not grow with read count. The 96 MB does.

**Wall time is a wash, not a win.** `BENCHMARKS.md` projected ~10% faster, measured against an `awk` stand-in converter; the real Rust converter is fast enough that the serial phase being hidden is nearly free. That projection is kept in the plan directory and explicitly superseded rather than quietly deleted. Conversion does come off the critical path, but the fan-out's thread handoffs cost about what it saves.

## Byte-identity, by construction

<details>
<summary>Details</summary>

`convert_fastq_impl` and `convert_fasta_impl` are split into a file wrapper plus a sink-generic record loop (`convert_fastq_into` / `convert_fasta_into`), and every other path drives the *same* loop. The conversion logic is not forked, so the paths cannot drift. Bismark writes its own `@PG`, so the one SAM field that records the input path never reaches the output. The 46 existing `convert` unit tests pass unchanged, which is the byte-freeze.

Verification is a 14-shape A/B: the same input through both arms, comparing `samtools view` records and the report. SE and PE across directional / pbat / non-directional, FastQ **and** FastA, `--gzip`, `-p 4`, `--multicore 2`, `--skip`/`--upto`, `--prefix`. All BAM- and report-identical, with the streaming arm's `--temp_dir` empty at the end of every one. The harness is committed at `plans/09112026_stream-converted/spikes/ab-run.sh`.

</details>

## Escape hatch and fallbacks

<details>
<summary>Details</summary>

`--no_stream_converted` forces the previous file path — for inspecting the converted reads, or bisecting. Runs that cannot stream fall back on their own and say why on STDERR:

| | Why |
|---|---|
| `--hisat2` | rejects a named pipe as its read input (exit 255 — it stats the file while sniffing the format). Its only disk lever stays `--gzip` |
| `--combined_index` models | still read the converted reads as files; see "What is deliberately not here" |
| `--rammap_subprocess` | the external binary reads from a path, and whether it accepts a FIFO is unverified |
| a `--temp_dir` that will not host a FIFO | probed at start-up with a real `mkfifo` |
| `--illumina_5base` | aligns to the *unconverted* genome, so it writes no converted reads at all — nothing to stream, nothing to explain |

**`--rammap`'s default in-process backend uses no intermediate at all**, not even a pipe. It aligns inside the Bismark process, so pointing it at a temp file meant the converter serialised records to disk purely so the same process could parse them back. It now receives the converted reads in memory. `InProcessAlignerStream` was already generic over `BufRead`, so `inprocess.rs` needed no change; a `ChannelReader` over the converter's existing bounded channel replaces the file, and the fan-out is shared with the FIFO path rather than duplicated. The channel deliberately carries **bytes, not parsed records**: records would also skip the format-and-parse round trip, but they would fork the conversion loop, and an unforked loop is what makes byte-identity structural. A test pins the in-memory bytes equal to what the file path writes.

`--gzip` is inert on the streaming path, since there is no file to compress. That costs nothing: `--gzip` has never reached an output file in either implementation — in the Rust aligner it feeds only `ConvertOptions`, and in Perl its only other use is the `.temp.N` subset *names*. The BAM is BGZF regardless and the `--unmapped`/`--ambiguous` FastQ are gzipped unconditionally in both.

`--multicore N` still writes its per-chunk input subsets. Only the converted reads are streamed — the chunking path is untouched, per the closing comment on #1091.

</details>

## Three structural changes worth reviewing closely

<details>
<summary>`AlignerStream::spawn` is split into `spawn_unprimed` + `prime`.</summary>

It previously spawned the child *and* read through to the first alignment record in one step. On a FIFO that read cannot complete until the writer is let past its blocking open-for-write, which happens only once a child opens the pipe — so priming instance 0 before instance 1 exists stalls the shared conversion pass on instance 1's unconsumed channel and deadlocks the run. The two halves are newtype-separated (`UnprimedAlignerStream`) because an unprimed stream's `current: None` is indistinguishable from EOF and would silently drop every alignment. The file path is unchanged: `spawn` is the two chained, and the eight combined-index drivers were not touched.

</details>

<details>
<summary>A failed instance is now reported instead of hanging the run.</summary>

The fan-out is bounded, so a consumer that never opens its pipe fills its channel and stalls the conversion pass it *shares* with the other instances; the siblings then block mid-merge on children that are alive but will never receive another read, and nothing times out. `await_readers` sits between the spawn and prime loops and waits for every pipe to find its reader, polling each child rather than using a deadline (an aligner loading a mammalian index can legitimately take minutes to open its reads). Only a **non-zero** exit before opening is fatal; a clean exit without reading is the shape of this repo's own fake aligners, and treating it as failure broke eight paired-end tests.

</details>

<details>
<summary>Teardown releases writers that are still parked.</summary>

A child that exits without ever opening leaves its writer blocked in `open`, and joining it never returns — not hypothetical, the fake `bowtie2` in `aligner_cli.rs` prints SAM records without reading a byte and deadlocked that whole suite. Each writer publishes its state once its open returns, and teardown opens and discards any pipe still parked. That state is an `AtomicU8` rather than a channel, because **two** readers need it (the handshake above and the teardown) and a one-shot channel gives the answer to whichever asks first; it is paired with `JoinHandle::is_finished()` so a writer that panicked before publishing anything is handled too.

</details>

## What the review changed

<details>
<summary>Details</summary>

A `/code-review` pass at high effort produced five findings. Four are fixed; the fifth turned out to have a better answer than the one proposed.

- **HIGH** — the silent hang above, now `await_readers`.
- Each 256 KiB block was deep-copied once per consumer (four copies per block for non-directional PE). Now `Arc<[u8]>`, so fan-out is a refcount bump.
- `summary()` reported `--illumina_5base` as writing converted files to the temp dir. It writes none; the line now has a third state.
- `try_recv().is_ok()` conflated "disconnected" with "never opened". Subsumed by the `AtomicU8` change.
- The review proposed an automatic per-chunk fallback to files when `mkfifo` fails on a long name. **The better fix was to stop the name growing**: these are intermediates with exactly one consumer and nothing functional reads their names, so the stem is capped at 64 bytes (truncated on a char boundary) and `ENAMETOOLONG` becomes unreachable. No fallback path needed.

Capping alone would have been worse than the bug. `--multicore N` runs its chunks concurrently in one process, so the pid and slot numbers repeat, and the only difference between chunk 0's pipe name and chunk 1's is `.temp.<i>` — at the **end** of the basename, exactly where the cap truncates. Both would have produced one path, and since `mkfifo` unlinks before creating, the second would have deleted the first's live pipe and both would have written into one: interleaved reads, wrong alignments, no error. Path uniqueness now comes from a process-wide set counter, leaving the name purely descriptive.

Worth flagging for reviewers: **neither that collision nor the fake-aligner hang could have been caught by the A/B matrix.** Real aligners always open their inputs, and the fixture names are too short to truncate. The container matrix proves byte-identity on realistic input; it does not probe edges. Each of those now has a unit test that asserts its *premise* before its conclusion, so it cannot quietly stop testing anything.

</details>

## What is deliberately not here

<details>
<summary>Details</summary>

**Combined-index streaming.** I checked all eight combined-index drivers for the thing that would actually block it — a converted file read more than once, which a pipe cannot serve — and none of them do it; every combined-index converted file has exactly one reader. So this is deferred on cost and risk, not feasibility: eight more drivers for an opt-in mode, two further conversion cores (the tagged-interleaved converters need the same split), and `--combined_index_sequential`'s whole purpose is minimising peak RSS by making pass 1's aligner exit before pass 2 spawns, which deserves its own design rather than a copy-paste. `process_se_chunk_combined` is the natural starting point: one pipe, one reader, no fan-out.

Worth noting who that leaves out: the mode guide steers disk-constrained non-directional users to `--combined_index_single_pass` or `--combined_index` (parallel) precisely because those have no spill — which means the converted reads are 100% of their scratch, and they gain nothing from this PR.

</details>

# Benchmarks

## Full-scale validation — all three claims pass

Ran [`FULLSCALE.md`](https://github.com/ewels/Bismark/blob/rust/stream-converted/plans/09112026_stream-converted/fullscale/FULLSCALE.md) on a 32-core / 124 GB machine: **54 benchmark runs, strictly one at a time**, `SRR24766921` (mouse RRBS PE, 51 bp) against Ensembl 112 GRCm39.

| | claim | result |
|---|---|---|
| **C1** | output byte-identical | ✅ **26 arm-pairs, 0 differing** |
| **C2** | converted temp files gone | ✅ **0 bytes** vs up to 6.38 GiB |
| **C3** | no wall-time regression | ✅ **never slower**, −0.01 % to −1.50 % |

---

### Disk, CPU, memory and output over one full-scale run

Directional PE, 10M pairs, ~49 minutes per arm. Blue is streaming (the new default), red is `--no_stream_converted`.

disk, CPU, memory and output

Four things in one picture:

- **Disk** — the file arm puts 3.19 GiB of converted reads down in the first 33 seconds and *holds it for the entire run*. Streaming never writes a byte.
- **CPU** — the file arm runs at 800 % for 29 minutes, then one strand hypothesis finishes and it spends the last 20 minutes at 400 %. Streaming holds a flat ~650 % and both instances finish together. Same total CPU either way (18,990 vs 18,993 CPU-seconds).
- **Memory** — streaming is **flat at 8.73 GiB**; the file arm **climbs to 12.05 GiB**, with another ~4 GiB of page cache above it (dashed) for its own converted files.
- **Output** — identical, as it must be.

---

### Disk: the headline

converted bytes on disk

Zero, not "small". Peak sampled *while the run is live* — both arms clean up at the end, so measuring afterwards reports zero for both and proves nothing.

<details>
<summary><b>The case this feature exists for — a <code>--temp_dir</code> that cannot hold the converted reads</b></summary>

<br>

`FULLSCALE.md` §7.2 notes this is *"the whole point of the feature, and nothing has demonstrated it yet"*. A 512 MiB tmpfs as `--temp_dir`, against a converted set needing roughly twice that:

| arm | exit | records |
|---|---|---|
| files | **1** — `I/O error: No space left on device (os error 28)` | 0 |
| streamed | **0** | 2,398,276 |

The file arm dies. The streamed arm completes with a full result set.

</details>

---

### Speed

wall time

Plotted relative to each group's files median, because the effects are under 2 % and absolute bars render both arms as identical rectangles.

| shape | scale | n | delta |
|---|---|---|---|
| `-p 2` | 2M | 4 | **−1.50 %** |
| `-p 4` | 2M | 5 | −0.66 % |
| `--non_directional` | 2M | 5 | −0.65 % |
| `--non_directional` | 10M | 3 | −0.55 % |
| `--multicore 2` | 10M | 1 | −0.50 % |
| `-p 4` | 10M | 4 | −0.43 % |
| `--multicore 2` | 2M | 5 | −0.01 % |

**Negative or zero in all seven groups.** The honest headline is *wall-neutral to slightly faster* — but the pre-merge concern was a **regression** from coupling the aligner instances, and there is no evidence of one.

<details>
<summary><b>How hard these numbers can be pushed (please read before quoting a percentage)</b></summary>

<br>

The `-p 4` 10M group has **overlapping distributions**: streamed 2937.2 · 2943.2 · 2946.8 · 2974.4 s against files 2938.9 · 2955.7 · 2960.0 · 2961.1 s. The per-rep spread exceeds the 12.9 s median difference. **Four reps do not establish a 0.43 % effect there.**

What the matrix *does* support is the claim C3 makes — not slower. Every median across seven independent groups is negative, and the two tightest (`-p 2` at −1.50 %, non-directional 2M at −0.65 %) have no overlap at all. A consistent sign across seven groups is strong; the magnitude on any single group is not.

The `--multicore 2` dead heat at −0.01 % is a useful control: fully interleaved distributions are what a genuine null looks like, which argues the rest isn't the harness quietly favouring an arm.

C1 and C2 are exact comparisons and carry no such caveat.

</details>

<details>
<summary><b>The instance-asymmetry question (#1120 D-6) — closed</b></summary>

<br>

`BENCHMARKS.md` closed on this as the known measurement gap: a tee couples the aligner instances, so a fast one can be throttled behind a slow one, and no fixture ran long enough to show whether that costs anything.

On directional 10M:

| | CPU profile | instances finish |
|---|---|---|
| **files** | 800 % until t=1761 s, then **400 % for the remaining 1194 s** | 20 minutes apart |
| **streamed** | flat **645 %** throughout | together, t=2944 s |

The coupling is **real and large** — with files the strand hypotheses fully decouple and the machine spends the last 40 % of the run at half utilisation. Streaming paces the fast instance to the slow one, which is exactly the throttling that was the concern.

**And it costs nothing.** Total CPU 18,990 vs 18,993 CPU-seconds; the streamed arm finished 8.9 s sooner. Wall time is set by the slowest instance in both arms — streaming doesn't slow that instance, it declines to let the fast one race ahead and then idle. Same work, spread evenly instead of front-loaded.

**`CHANNEL_DEPTH` was not touched and does not need raising on this evidence.**

</details>

---

### Memory

peak memory

Peak **allocation** (page cache excluded), lower streamed in six of seven groups and tied in the seventh. Biggest gap is `--multicore 2` at 10M: 14.95 vs 18.09 GiB — two concurrent chunks each reading back their own converted files is where the file path costs most.

<details>
<summary><b>A measurement trap worth knowing about</b></summary>

<br>

cgroup `memory.current` counts page cache, so the file arm's *total* runs 6+ GiB above the streamed arm purely because its own converted files are cached. Read unqualified, that would trip `FULLSCALE.md` §2's tripwire — *"if peak RSS moves by more than a few hundred MB, something is wrong"* — and manufacture a failure that isn't one.

The harness records `memory.stat anon` separately for this reason. Every memory figure quoted here is allocation, not cache. The dashed lines in the first figure are the totals, for comparison.

</details>

---

<details>
<summary><b>Method</b></summary>

<br>

- **One benchmark at a time**, behind a lockfile. Never two arms, never two shapes, nothing else on the box.
- **Each arm in its own container**, so cgroup v2 counters (`memory.current`, `memory.peak`, `cpu.stat`, `pids.current`) cover the whole process tree including every `bowtie2` child, starting from zero — not the parent's RSS.
- **Page cache warmed identically before every run** (13 GB index into 124 GB RAM), so no run is penalised for faulting it in.
- **Arm order alternated between reps** — a fixed order hands any residual first-run penalty to the same arm every time.
- **~1 s time series per run**: converted bytes, whole temp dir, output, memory (total and anonymous), CPU, threads. The sample loop costs ~16 ms per tick, so each sample's timestamp is honest — an earlier version spent over a second per tick, which is fine for a peak and fatal for a curve whose whole content is *when* the bytes appear.
- **C1 compares `samtools view` record bodies**, not raw BAM: the header's `@PG CL:` line is the verbatim argv and differs between arms by design. Reports compared with absolute paths filtered.

Harness, raw time series, per-run plots and the full record: [`fullscale/`](https://github.com/ewels/Bismark/tree/rust/stream-converted/plans/09112026_stream-converted/fullscale) · summary in [`BENCHMARKS.md` §3](https://github.com/ewels/Bismark/blob/rust/stream-converted/plans/09112026_stream-converted/BENCHMARKS.md).

</details>

<details>
<summary><b>Section 7 — the edge cases, all pass</b></summary>

<br>

**7.1 — long sample names under `--multicore`.** FIFO names cap the stem at `NAME_STEM_CAP` (64 bytes) and uniqueness comes from a process-wide counter rather than the name, so a collision would surface as interleaved or truncated output rather than an error. Checked by comparing alignment columns against a short-named control, not by exit code:

| run | basename | exit | records |
|---|---|---|---|
| control | 26 chars | 0 | 2,398,276 |
| long name | **203 chars**, `--multicore 4` | 0 | 2,398,276 |

Alignment-identical. The fix that had unit tests but had never run at scale holds.

**7.3 — thread and FD counts**, under the widest fan-out (`--non_directional --multicore 2`, 8 pipes concurrently): peak **73** open FDs against a 65,535 soft limit (0.1 %), 35 threads, 9 processes. Three orders of magnitude of headroom.

</details>

<details>
<summary><b>What did not reproduce, and what is not covered</b></summary>

<br>

**A 2.4 % non-directional regression — retracted.** The first 10M non-directional run had the streamed arm 72 s slower, and a plausible mechanism was available (four consumers sharing two converted streams, less slack in the fan-out). With three reps the same shape is **0.65 % faster**; the original run was the slowest of three, sitting 87 s above the next streamed rep against a files spread of 27 s. An outlier. It is left in [`RESULTS.md`](https://github.com/ewels/Bismark/blob/rust/stream-converted/plans/09112026_stream-converted/fullscale/RESULTS.md) §4.3 → §4.7 rather than edited away, because the story survived on plausibility rather than evidence until repetition killed it.

**Not covered:**
- **Combined-index modes**, `--hisat2` and `--rammap_subprocess` keep files by design and are unchanged by this PR (`PROGRESS.md` D-11).
- **One dataset, one genome, one machine.** The human WGBS accession `SRR24827373` named in `FULLSCALE.md` §3 was not run.
- **A clean A/B is not evidence about edges.** Three defects in this work were invisible to the 14-shape container matrix; §7 probes three specific edges, it does not exhaust them.

</details>

<details>
<summary><b>One defect in the brief: <code>-p 1</code> is not a runnable shape</b></summary>

<br>

`FULLSCALE.md` §4 lists *"Directional PE, `-p 1` — the fan-out has the least slack here"* as one of four shapes to cover. Bismark rejects it:

```
error: Please select a value for -p of 2 or more!
```

`aligner/options.rs:165`, matching Perl 7993-8007. Both arms exited 1 in about a second. Substituted **`-p 2`**, the least-slack configuration that actually runs.

Worth noting the brief's reasoning was also inverted by the result: it nominated the fewest-threads shape as the one where streaming was *most likely to hurt*. It is the shape where streaming helps **most** (−1.50 %). The ordering is consistent — the fewer threads each aligner instance has, the more streaming wins.

</details>

---

**Recommendation: merge.** No knob needs turning.


ewels:
## The reviewer agents this workflow depends on aren't in the repo

`PROGRESS.md` in this PR carries two steps I couldn't complete as written:

| # | Step | Note in the table |
|---|---|---|
| 7 | Agent review (dual `plan-reviewer`) | *"Repo convention for this workflow; not run — ask before spawning agents"* |
| 15 | Verify (dual `code-reviewer`) | *"Ask before spawning"* |

Neither agent type exists in a fresh clone. There is no `.claude/` directory in the repository at all, so `plan-reviewer`, `code-reviewer` and `plan-manager` must be defined on your own machine rather than version-controlled alongside the work they produce.

That matters more than it looks, because the workflow is not incidental — it is the house style, and its output is committed:

| Artefact | Plan directories containing it |
|---|---|
| `CODE_REVIEW_A.md` + `CODE_REVIEW_B.md` | 25 |
| `PLAN_REVIEW_A.md` + `PLAN_REVIEW_B.md` | 24 |
| `COVERAGE.md` (the `plan-manager` audit) | 23 |

So the `plans/` tree documents a process in detail, cites its verdicts as evidence, and treats dual review as a gate — while the thing that performs it ships with nobody but you. A contributor reading `plans/07292026_local-mapq-denominator/PROGRESS.md` can see that "both reviewers found the false premise" and cannot reproduce that step. The same is true of the instruction in this PR's own `PROGRESS.md`.

**Suggestion: commit the agent definitions as `.claude/agents/*.md`.** Claude Code discovers project-level agents there automatically, so they would be available to anyone who clones the repo, they would version alongside the conventions they encode, and changes to a reviewer's brief would show up in `git log` like anything else. A short section in `CONTRIBUTING.md` pointing at them would close the loop, since the plans tree currently assumes knowledge that lives outside it.

Worth saying plainly: I have no idea whether those definitions contain anything you would rather not publish. If they do, a line in `CONTRIBUTING.md` saying the dual-review steps are maintainer-only would at least stop the `plans/` files reading as instructions a contributor has failed to follow.

**What I did instead for this PR:** ran the equivalent review with the tooling that is available and folded the findings in. Details in the next comment. Nothing here blocks the PR; it is a note about the gap between what the plan files prescribe and what the repository provides.

ewels:
## Code review results

Ran `/code-review` at high effort against this PR in place of the missing dual `code-reviewer` step. Five findings; four fixed, one partly. Fixes are in `444c0c6`, with `1014221` covering a defect I found separately while waiting.

### HIGH — an instance that died during start-up hung the run

The fan-out is bounded (`CHANNEL_DEPTH * BLOCK_BYTES`, about 1 MiB per consumer), so a consumer that never opens its FIFO fills its channel and then stalls the conversion pass it *shares* with the other instances. The siblings go on to block mid-merge on children that are alive but will never receive another read, and nothing times out — the run stops dead with no message, after appearing to make progress. Reachable via a corrupt-but-present index, an OOM-killed child, or a wrapper that rejects the argv.

This is a regression in failure behaviour specifically, not in correct-run behaviour: the file path fails loudly here because each instance reads its own file, so a dead one surfaces as a non-zero exit. `drain_unopened` could not help, since it only runs from `finish`/`Drop` and neither is reached.

Fixed with an `await_readers` handshake between the spawn and prime loops. It polls each child rather than using a deadline — an aligner loading a mammalian index can legitimately take minutes before it opens its reads, so any fixed timeout would be flaky or useless.

**Only a non-zero exit before opening is fatal.** My first attempt treated any early exit as fatal and broke eight paired-end tests, because this repo's fake aligners print SAM records without consuming a byte. A clean exit without reading has produced whatever it was going to produce, so the run continues and teardown releases the writer. Recorded as D-12, because the container A/B *cannot* catch that regression — real aligners always open their inputs, so the 14-shape matrix passed even against the broken version. The unit tests are the only guard.

### A defect the fix then introduced, found and fixed

Adding the handshake gave the writer-open signal two readers — the handshake and the teardown — and it was a one-shot channel, so whichever asked first consumed it and left the other guessing. Teardown then opened pipes it should have skipped. In production that survived only by accident, through the `Disconnected` case of an unrelated fix in the same batch.

It is now an `AtomicU8` both sides poll without consuming, paired with `JoinHandle::is_finished()` so a writer that panicked before publishing any state is handled too — which the channel version silently was not. D-13.

Worth saying how it surfaced: the full workspace suite appeared to be running slowly, and was in fact hung. `ps` showed 0.0% CPU, and `sample` put the block inside the test I had just added, which attached a reader and joined it — a reader that can never finish, precisely because the stalled consumer starves it. The test was reproducing the bug on itself.

### The rest

| Finding | Disposition |
|---|---|
| `try_recv().is_ok()` conflated "disconnected" with "never opened", so a writer that died before signalling turned teardown into a hang | Fixed, subsumed by the `AtomicU8` change |
| Each 256 KiB block was deep-copied once per consumer — four copies per block for non-directional PE | Fixed: `Arc<[u8]>`, so fan-out is a refcount bump. This is the cost the PR description blames for the lost speedup |
| `summary()` reported `--illumina_5base` as writing converted files to the temp dir, when it writes none | Fixed: third state |
| A per-chunk `mkfifo` failure aborts instead of falling back to files | **Partly.** See below |

### Left open, deliberately

A per-chunk `mkfifo` failure still aborts the run where the docs promise an automatic fallback. It is genuinely reachable after a successful start-up probe: the probe uses a short fixed name, but the real FIFO name adds `.s<slot>.<pid>.fifo` — and `.temp.<i>` again under `--multicore` — so a long sample basename that worked as a file can exceed 255 bytes as a pipe. A temp dir that fills or goes read-only mid-run does the same.

The error now names `--no_stream_converted`, which unblocks anyone who hits it. The proper fix means restructuring `process_se_chunk`/`process_pe_chunk` so the file-path body is callable after a streaming start fails, which felt like the wrong trade against a condition needing a ~240-character basename. Recorded as an Open item in `PROGRESS.md` — happy to do it properly if you would rather.

### Verification after the fixes

14-shape A/B byte-identical again (SE and PE across directional / pbat / non-directional, FastQ and FastA, `--gzip`, `-p 4`, `--multicore 2`, `--skip`/`--upto`, `--prefix`), full workspace suite green with zero failures, 1516 lib tests in 1.3 s, `fmt` and `clippy -D warnings` clean. Three new unit tests, including the clean-exit case my first attempt got wrong.

ewels:
## Two follow-ups, both from your review rather than the automated one

### The `mkfifo` name-length finding is gone, not deferred

I had this wrong. The earlier comment said a per-chunk fallback to files was the proper fix and left it open. The right question was why the pipe's name matters at all: these are intermediates with exactly one consumer, and nothing functional reads the name. The aligners sniff content; only the format extension matters, and that stays last.

So the stem is capped at 64 bytes, truncated on a character boundary so a non-ASCII sample name cannot panic. `ENAMETOOLONG` is unreachable, and no fallback path is needed. The Open item in `PROGRESS.md` is removed.

**Capping alone would have been worse than the bug it fixed**, and it is worth recording why. `--multicore N` runs its chunks concurrently in one process, so the pid and the slot numbers repeat across chunks. The only remaining difference between chunk 0's pipe name and chunk 1's is `.temp.<i>` — which sits at the **end** of the basename, exactly where the cap truncates. Both chunks would have produced the same path, and since `mkfifo` unlinks before it creates, the second would have deleted the first's live pipe and both would have written into one: interleaved reads, wrong alignments, no error anywhere.

Path uniqueness now comes from a process-wide set counter instead of the name, which leaves the name free to be purely descriptive. The regression test asserts the collision premise first, so it cannot quietly stop testing anything if the truncation ever changes.

Worth being blunt about the gap this exposes: **the 14-shape A/B matrix could not have caught it**, because its fixture names are short enough that nothing truncates. Same lesson as the fake-aligner case earlier in this PR. The container matrix proves byte-identity on realistic inputs; it does not probe the edges, and the unit tests have to.

### `--rammap`'s in-process backend now uses no intermediate at all

Streaming was the wrong fix for this one. That backend aligns *inside* the Bismark process, yet its converted reads still went to disk: the converter serialised records to a temp file so that the same process could parse them straight back. A FIFO would have removed the disk and kept the syscalls, plus a writer thread per instance.

`InProcessAlignerStream` was already generic over `BufRead`, so `inprocess.rs` needed **no change**. A `ChannelReader` over the converter's existing bounded channel replaces the file, and the converter fan-out is shared with the FIFO path rather than duplicated.

One deliberate choice: the channel carries **bytes, not parsed records**. Passing records would also skip the format-and-parse round trip, but it would fork the conversion loop, and an unforked loop is exactly what makes byte-identity hold by construction rather than by argument. A test pins the in-memory bytes equal to what the file path writes, for every consumer, from one conversion pass. If that round trip ever shows up in a profile, it can be revisited on its own terms.

`--no_stream_converted` still forces files here too, and `--rammap_subprocess` keeps them, since whether the external binary accepts a FIFO is unverified. The routing and `resolve_stream_converted` read the same `inprocess_rammap_selected` predicate, so the printed decision and the path taken cannot disagree.

### Verification

14-shape A/B byte-identical, workspace suite green across 82 test binaries with zero failures, 1528 lib tests with `--features rammap-inprocess`, and `fmt` plus `clippy -D warnings` clean in both feature states.

Combined-index streaming remains the one deferred item, for the reasons in `PROGRESS.md` (D-11).

